PR #233, opened 2017-12-06
Zoo updates with details for each model

### LINKED ISSUE #193: Model zoo need high level description of models and s3 links for archive
https://github.com/awslabs/mxnet-model-server/blob/master/docs/model_zoo.md

1. We should have 1 paragraph summary of all the models we have in model zoo. Answer on what dataset it is trained, what is the input and output, example usecase.
2. Provide S3 links for each .model archive. Better if we can give command like below for all the models
mxnet-model-server \
  --models squeezenet=https://s3.amazonaws.com/model-server/models/squeezenet_v1.1/squeezenet_v1.1.model
3. Link this from main README file? from the point where we talk about serving that more models are available. https://github.com/awslabs/mxnet-model-server/blob/master/README.md#serve-a-model

aaronmarkham:
Added example invocations for each as well as a link to the scholarly article about it and a short extract from the abstract.

Fixes #193

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ The predictor endpoint will return a prediction response in JSON. It will look s
   ]
 }
 ```
+Other models can be downloaded from the [model zoo](docs/model_zoo.md), so try out some of those as well.
 
 Now you've seen how easy it can be to serve a deep learning model with MMS! [Would you like to know more?](docs/server.md)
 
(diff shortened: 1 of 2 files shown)


lupesko:
@aaronmarkham please see my comments. I'm not so sure we want this change in the way it is now.

Issue #241, opened 2017-12-28
Cannot start mxnet-model-server on non-default port.


treske:
Using the --port option for mxnet-model-server fails with the following error:
Failed to start model serving host: Flask handler failed to start: port must be an integer

Steps to recreate:

```bash
mkdir -p ~/env/mxnet-model-server-port-issue 
python3 -m venv ~/env/mxnet-model-server-port-issue
source ~/env/mxnet-model-server-port-issue/bin/activate
pip install mxnet-model-server    
cd ~/env/mxnet-model-server-port-issue
mxnet-model-server --models squeezenet=https://s3.amazonaws.com/model-server/models/squeezenet_v1.1/squeezenet_v1.1.model --log-file ~/env/mxnet-model-server-port-issue/mms.log --log-level DEBUG --host 127.0.0.1 --port 8090
```

One way to fix would be to change line 135 in mms/mxnet_model_server.py

from
self.port = self.args.port or 8080
to
self.port = int(self.args.port) or 8080

lupesko:
Thanks for finding the issue.
Feel free to contribute the fix, or I can push a fix in a day or two.

yuruofeifei:
Fixed at #240

PR #240, opened 2017-12-28
Add onnx support

yuruofeifei:
@nswamy @sandeep-krishnamurthy

PR #242, opened 2017-12-28
Update mxnet_model_server.py

treske:
Fix issue where Flask handler fails to start when using non-default port. See #241

### DIFF
--- a/mms/mxnet_model_server.py
+++ b/mms/mxnet_model_server.py
@@ -132,7 +132,7 @@ class MMS(object):
         """
         try:
             # Port
-            self.port = self.args.port or 8080
+            self.port = int(self.args.port) or 8080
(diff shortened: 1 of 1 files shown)


mms/mxnet_model_server.py
```diff
@@ -132,7 +132,7 @@ def _arg_process(self):
         """
         try:
             # Port
-            self.port = self.args.port or 8080
+            self.port = int(self.args.port) or 8080
```
yuruofeifei: Please do none check before conversion.

yuruofeifei:
Hi @treske , thanks for the PR.
Please do none check before conversion.

yuruofeifei:
Thanks for submitting the PR, it is fixed at #240

Issue #278, opened 2018-01-16
engine is hardcoded in the manifest


aaronmarkham:
I'm pretty sure when I updated I have > 1.0 of MXNet. However, my manifest.json says my export used v0.12. I don't think that's true.
Here's the hardcoding line:
https://github.com/awslabs/mxnet-model-server/blob/c86255338edac3e4c706e9ec127fdfb17eec60c3/mms/export_model.py#L189

Issue #274, opened 2018-01-12
CloudWatch metrics feature issues


lupesko:
Users of CloudWatch feature should be able to:
- Monitor all metrics per host
- Monitor all metrics per model (across hosts)

Currently they cannot. Let's please make a better use of the [CloudWatch BOTO API] (http://boto3.readthedocs.io/en/latest/reference/services/cloudwatch.html) so:
- Use Dimensions to capture the ModelName and the HostName (two different dimensions)
- Use StatisticValues to capture statistical info such as latencies (*_latency)
- We’re missing a postprocess_latency – let’s please add it
- Let’s please use the namespace “ModelServer/”
- Let's please differentiate 4XX errors from 5XX errors

@jesterhazy @nswamy @yuruofeifei for feedback and suggestions

jesterhazy:
also fix typo in mxnet-model-server.py mode_name -> model_name

jesterhazy:
latencies are currently measured in seconds, which isn't too helpful. let's change to milliseconds.

jesterhazy:
how is memory metric calculated?

`memory for last 30 seconds is 0.006215, writing to AWS CloudWatch...`

this is with 16GB total memory, mms python process using 106MB.

jesterhazy:
how are latencies calculated? it looks like a sum of latencies over last reporting window, which isn't useful. we should discuss correct reporting for these values.

if we report as individual data points (a list of latencies per CW call), then users can use standard CW aggregation to get useful stats like min/max/mean/med and percentiles.

if we want to report fewer datapoints, then we could collect our own summary stats, but then we'd need to choose some useful summary metrics, and report each of them separately. I prefer the separate data points approach. See https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/PublishMetrics.html for some background

jesterhazy:
+1 to separating 4XX 5XX errors

jesterhazy:
for the namespace, i think we should stay aligned with the product name. That would give us 'MXNetModelServer'

lupesko:
> latencies are currently measured in seconds, which isn't too helpful. let's change to milliseconds.

Agreed we need to report ms, but beyond that we should use proper units for every metric we're reporting. CloudWatch API supports pretty extensive units:
[...]

Issue #239, opened 2017-12-21
Wrong custom service class definition when the name is sorted before 'MXNetVisionService'


windsorwho:
https://github.com/awslabs/mxnet-model-server/blob/f98a2763f71695a59a69974c367c6f8e43203373/mms/mxnet_model_server.py#L150

The assumption that "# The overrided class is the last one in class_defs" is wrong because 'class_defs' is a list returned by inspect.getmembers(), which returns a list of member classes names sorted by the name. 
https://docs.python.org/2/library/inspect.html#inspect.getmembers

So, using "mode_class_name = class_defs[-1].__name__" leads to wrong class name when the class name string is something smaller than the base class name.

Thanks.

lupesko:
@windsorwho good call. Did this cause an issue for you? if so - can you provide the example?
@yuruofeifei can you take a look?

yuruofeifei:
@windsorwho, thank you for pointing this out!

@lupesko, I will take a look!

PR #283, opened 2018-01-16
Improve cloudwatch integration, fix several issues.

yuruofeifei:
Improve cloudwatch integration, fix several issues.

#278 #274 #239

@lupesko @jesterhazy

### DIFF
--- a/mms/export_model.py
+++ b/mms/export_model.py
@@ -47,6 +47,8 @@ MODEL_SERVER_VERSION = 0.1
 MANIFEST_FILE_NAME = 'MANIFEST.json'
 MXNET_TYPE = 'mxnet'
 ONNX_TYPE = 'onnx'
+MXNET_ATTRS = 'attrs'
+MXNET_VERSION = 'mxnet_version'
 
 NO_MODEL_FILES_MESSAGE = '''
 No model files found in the model directory {}.
@@ -187,13 +189,17 @@ def generate_manifest(symbol_file, params_file, service_file, signature_file, mo
     manifest["Model"]["Description"] = model_name
     manifest["Model"]["Model-Name"] = model_name
     manifest["Model"]["Model-Format"] = "MXNet-Symbolic"
-<<<<<<< HEAD
-    manifest["Engine"] = {"MXNet": 0.12}
-
-=======
-    manifest["Engine"]  = {"MXNet":mx.__version__}
     
->>>>>>> Fix
+    mxnet_version = mx.__version__
+
+    if os.path.exists(symbol_file):
+        symbol_json = json.load(open(symbol_file))
+        if MXNET_ATTRS in symbol_json:
+            if MXNET_VERSION in symbol_json[MXNET_ATTRS]:
+                mxnet_version = symbol_json[MXNET_ATTRS][MXNET_VERSION]
+
+    manifest["Engine"]  = {"MXNet": mxnet_version}           
+
     return manifest
 
 
--- a/mms/export_model.py
+++ b/mms/export_model.py
@@ -187,8 +187,13 @@ def generate_manifest(symbol_file, params_file, service_file, signature_file, mo
     manifest["Model"]["Description"] = model_name
     manifest["Model"]["Model-Name"] = model_name
     manifest["Model"]["Model-Format"] = "MXNet-Symbolic"
+<<<<<<< HEAD
     manifest["Engine"] = {"MXNet": 0.12}
 
+=======
+    manifest["Engine"]  = {"MXNet":mx.__version__}
+    
+>>>>>>> Fix
     return manifest
 
 
--- a/mms/metric.py
+++ b/mms/metric.py
@@ -13,6 +13,7 @@ import datetime
 import threading
 import os
 import warnings
+import socket
 
 from mms.log import get_logger
 
@@ -29,17 +30,31 @@ class Metric(object):
     """Metric class for model server
     """
     def __init__(self, name, mutex,
-                 namespace,
+                 model_name,
+                 unit,
+                 is_resource=False,
                  interval_sec=30,
                  update_func=None,
                  aggregate_method='interval_average',
                  write_to='log'):
         """Constructor for Metric class
 
+           Metric class will spawn a thread and report collected metrics to local disk 
+           or AWS cloudwatch between given intervals.
+
         Parameters
         ----------
         name : str
             Name of the metric
+        mutex: object
+            Lock between metric threads when they are writing local disk
+        model_name: str
+            Model name
+        unit: str
+            CloudWatch compatible unit
+        is_resource: boolean
+            Whether the metric is collecting host resource, 
+            if not we will add host, model_name dimensions
         interval_sec : int, optional
             Interval seconds between each data points, default 5 min
         update_func : func, optional
@@ -51,7 +66,7 @@ class Metric(object):
         """
         self.name = name
         self.interval_sec = interval_sec
-        self.namespace = namespace
+        self.model_name = model_name
 
         # Metrics within interval
         self.interval_datapoints_count = 0
@@ -64,6 +79,8 @@ class Metric(object):
         self.aggregate_method = aggregate_method
         self.mutex = mutex
         self.write_to = write_to
+        self.unit = unit
+        self.is_resource = is_resource
 
         # Setup cloudwatch handle
         if self.write_to == 'cloudwatch':
@@ -135,15 +152,28 @@ class Metric(object):
                 logger.info('Metric %s for last %s seconds is %f, writing to AWS CloudWatch...' %
                     (self.name, self.interval_sec, metric))
                 try:
+                    metric_data = {
+                        'MetricName': self.name,
+                        'Timestamp': utcnow,
+                        'Value': metric,
+                        'Unit': self.unit
+                    }
+                    if not self.is_resource:
+                        metric_data.update({
+                            'Dimensions': [
+                                {
+                                    'Name': 'host',
+                                    'Value': socket.gethostname()
+                                },
+                                {
+                                    'Name': 'model_name',
+                                    'Value': self.model_name
+                                }]
+                        })
                     self.client.put_metric_data(
-                        Namespace=self.namespace,
+                        Namespace='ModelServer',
                         MetricData=[
-                            {
-                                'MetricName': self.name,
-                                'Timestamp': utcnow,
-                                'Value': metric,
-                                'Unit': 'Count'
-                            },
+                            metric_data
                         ]
                     )
                 except Exception as e:
--- a/mms/metric.py
+++ b/mms/metric.py
@@ -17,6 +17,7 @@ import socket
 
 from mms.log import get_logger
 
+
 try:
     import boto3 as boto
 except ImportError:
@@ -25,6 +26,12 @@ except ImportError:
 
 logger = get_logger()
 
+MetricUnit = {
+  'ms': "Milliseconds",
+  'percent': 'Percent',
+  'count': 'Count',
+  'MB': 'Megabytes'
+}
 
 class Metric(object):
     """Metric class for model server
@@ -32,8 +39,8 @@ class Metric(object):
     def __init__(self, name, mutex,
                  model_name,
                  unit,
-                 is_resource=False,
-                 interval_sec=30,
+                 is_model_metric=True,
+                 interval_sec=60,
                  update_func=None,
                  aggregate_method='interval_average',
                  write_to='log'):
@@ -52,9 +59,8 @@ class Metric(object):
             Model name
         unit: str
             CloudWatch compatible unit
-        is_resource: boolean
-            Whether the metric is collecting host resource, 
-            if not we will add host, model_name dimensions
+        is_model_metric: boolean
+            Whether it is a metric for model eg. inference latency, requests count
         interval_sec : int, optional
             Interval seconds between each data points, default 5 min
         update_func : func, optional
@@ -71,6 +77,8 @@ class Metric(object):
         # Metrics within interval
         self.interval_datapoints_count = 0
         self.interval_metric_aggregate = 0.0
+        self.min_value = None
+        self.max_value = None
 
         # Metrics for the whole session
         self.total_datapoints_count = 0
@@ -80,7 +88,7 @@ class Metric(object):
         self.mutex = mutex
         self.write_to = write_to
         self.unit = unit
-        self.is_resource = is_resource
+        self.is_model_metric = is_model_metric
 
         # Setup cloudwatch handle
         if self.write_to == 'cloudwatch':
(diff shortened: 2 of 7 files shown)


mms/export_model.py
```diff
@@ -180,7 +180,7 @@ def generate_manifest(symbol_file, params_file, service_file, signature_file, mo
     manifest["Model"]["Description"] = model_name
     manifest["Model"]["Model-Name"] = model_name
     manifest["Model"]["Model-Format"] = "MXNet-Symbolic"
-    manifest["Engine"]  = {"MXNet":0.12}
+    manifest["Engine"]  = {"MXNet":mx.__version__}
```
lupesko: hmm. So we're grabbing this off the locally installed MXNet - is this what we want to do? what happens if the user has an old version of MXNet?

mms/metric.py
```diff
@@ -29,7 +30,9 @@ class Metric(object):
     """Metric class for model server
```
lupesko: Can you please add proper docstring for this ctor and all input args?

mms/metrics_manager.py
```diff
@@ -33,21 +33,36 @@ def memory(metric_instance):
...
+
     timer = threading.Timer(intervalSec, disk, [metric_instance])
     timer.daemon = True
     timer.start()
 
+Unit = {
+  'ms': "Milliseconds",
+  'percent': 'Percent',
+  'count': 'Count',
+  'mb': 'Megabytes'
```
lupesko: Can we change this to MB, to differentiate from "m" which is milli.

mms/metrics_manager.py
```diff
@@ -33,21 +33,36 @@ def memory(metric_instance):
...
+
+def disk(metric_instance):
+    disk_usage = psutil.disk_usage('/').used / (1024 * 1024) # in MB
+    metric_instance.update(disk_usage)
+
     timer = threading.Timer(intervalSec, disk, [metric_instance])
     timer.daemon = True
     timer.start()
 
+Unit = {
```
lupesko: Why not have this in a separate file?

lupesko: Can we name this MetricUnit?

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
+                                                                model_name=model_name,
+                                                                unit=Unit['count'],
+                                                                aggregate_method='interval_sum',
+                                                                write_to=metrics_write_to)
+            MetricsManager.metrics['400error_metric'] = Metric('400error_metric', mutex,
+                                                                model_name=model_name,
+                                                                unit=Unit['count'],
+                                                                aggregate_method='interval_sum',
+                                                                write_to=metrics_write_to)
+            MetricsManager.metrics['request_metric'] = Metric('requests', mutex,
```
lupesko: Can't we use the same string key instead of having both "request_metric" and "requests"?
Same for all of the next metrics.

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
+
         MetricsManager.metrics['cpu_metric'] = Metric('cpu', mutex,
-                                                      namespace=namespace,
+                                                      model_name=None,
+                                                      unit=Unit['percent'],
+                                                      is_resource=True,
                                                       aggregate_method='interval_average',
                                                       write_to=metrics_write_to,
                                                       update_func=cpu)
         MetricsManager.metrics['memory_metric'] = Metric('memory', mutex,
```
lupesko: Is this process memory, or host memory? Can we clarify this with proper naming? e.g. host_memory or process_memory
Same applies to cpu and IO.

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
-                                                         namespace=namespace,
+                                                         model_name=None,
+                                                         unit=Unit['percent'],
+                                                         is_resource=True,
                                                          aggregate_method='interval_average',
                                                          write_to=metrics_write_to,
                                                          update_func=memory)
+        MetricsManager.metrics['disk_percentage_metric'] = Metric('disk_percentage', mutex,
```
lupesko: Is this DiskIO or overall IO?

mms/mxnet_model_server.py
```diff
@@ -156,15 +156,15 @@ def _arg_process(self):
...
             class_defs = self.serving_frontend.register_module(self.args.service or service_file)
-            
-            if len(class_defs) < 1:
-                raise Exception('User defined module must derive base ModelService.')
-            # The overrided class is the last one in class_defs
-            mode_class_name = class_defs[-1].__name__
+            class_defs = filter(lambda c: len(c.__subclasses__()) == 0, class_defs)
+
+            if len(class_defs) != 1:
+                raise Exception('There should be one user defined service derived from ModelService.')
```
lupesko: Can't there be more than one?

mms/serving_frontend.py
```diff
@@ -370,8 +370,8 @@ def ping_callback(self, **kwargs):
             for model in self.service_manager.get_loaded_modelservices().values():
                 model.ping()
         except Exception:
-            if 'error_metric' in MetricsManager.metrics:
-                MetricsManager.metrics['error_metric'].update(metric=1)
+            if '500error_metric' in MetricsManager.metrics:
```
lupesko: Can we call this 5XX and not 500? Same for 4XX vs 400.

mms/export_model.py
yuruofeifei: We are using installed(local) mxnet to do model save, I think it makes sense to put it here.

mms/metric.py
yuruofeifei: updated.

mms/metrics_manager.py
yuruofeifei: I think the unit map and metric helper function is only relevant to this file.

yuruofeifei: changed

mms/metrics_manager.py
yuruofeifei: Changed.

mms/metrics_manager.py
yuruofeifei: changed

mms/metrics_manager.py
yuruofeifei: changed.

mms/metrics_manager.py
yuruofeifei: DiskIO

mms/mxnet_model_server.py
yuruofeifei: I think one service class make senses, if it is more than 1, we will not know which one to start the server.

mms/serving_frontend.py
yuruofeifei: changed.

mms/export_model.py
jesterhazy: for onnx models this might make sense, since we are creating the symbol and params file with current version of mxnet. but for user exporting (packaging) existing mxnet model, the current version might not match the actual model files, so we would be creating incorrect manifest. Can we parse this from the model files instead?

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
+                                                                write_to=metrics_write_to)
+            MetricsManager.metrics['requests'] = Metric('requests', mutex,
+                                                              model_name=model_name,
+                                                              unit=MetricUnit['count'],
+                                                              aggregate_method='interval_sum',
+                                                              write_to=metrics_write_to)
+            MetricsManager.metrics['overall_latency'] = Metric('overall_latency', mutex,
+                                                                      model_name=model_name,
+                                                                      unit=MetricUnit['MB'],
+                                                                      aggregate_method='interval_average',
```
jesterhazy: interval_average (and total_average) methods mean we are reporting snapshot of average value for a time window (or cumulative). it will be impossible for users to aggregate these values across hosts or models in cloudwatch. we need to report these as CW 'Statistics', not as individual data points that represent the average.

also, since reporting as Statistics means CW can't calculate percentiles (e.g. TP99s), we will need to calculate those and report them in the same statistics object. we can do that as separate PR, as long as we get the statistics structure right here.

mms/metric.py
```diff
@@ -29,17 +30,31 @@ class Metric(object):
...
         name : str
             Name of the metric
+        mutex: object
+            Lock between metric threads when they are writing local disk
+        model_name: str
+            Model name
+        unit: str
+            CloudWatch compatible unit
+        is_resource: boolean
+            Whether the metric is collecting host resource,
```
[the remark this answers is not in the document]
yuruofeifei: Agree, will improve.

mms/metric.py
```diff
@@ -135,15 +152,28 @@ def start_recording(self):
...
+                                    'Value': socket.gethostname()
+                                },
+                                {
+                                    'Name': 'model_name',
+                                    'Value': self.model_name
+                                }]
+                        })
                     self.client.put_metric_data(
-                        Namespace=self.namespace,
+                        Namespace='ModelServer',
```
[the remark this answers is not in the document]
yuruofeifei: I'm not sure if we want to change to MXNetModel server @lupesko

mms/metrics_manager.py
```diff
@@ -33,21 +33,36 @@ def memory(metric_instance):
...
+def disk_percentage(metric_instance):
     disk_usage = psutil.disk_usage('/').percent / 100.0
     metric_instance.update(disk_usage)
 
+    timer = threading.Timer(intervalSec, disk_percentage, [metric_instance])
+    timer.daemon = True
+    timer.start()
+
+def disk(metric_instance):
+    disk_usage = psutil.disk_usage('/').used / (1024 * 1024) # in MB
```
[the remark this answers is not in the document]
yuruofeifei: percentage is also kept above

mms/metrics_manager.py
yuruofeifei: If we want to migrate metrics entirely to CW, the concern is we will lose the usage of other reporting method. I'm not sure. @aaronmarkham @lupesko

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
+                                                                      model_name=model_name,
+                                                                      unit=MetricUnit['MB'],
+                                                                      aggregate_method='interval_average',
+                                                                      write_to=metrics_write_to)
+            MetricsManager.metrics['inference_latency'] = Metric('inference_latency', mutex,
+                                                                        model_name=model_name,
+                                                                        unit=MetricUnit['ms'],
+                                                                        aggregate_method='interval_average',
+                                                                        write_to=metrics_write_to)
+            MetricsManager.metrics['pre_latency'] = Metric('preprocess_latency', mutex,
```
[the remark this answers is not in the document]
yuruofeifei: MXNet symbol is lazy evaluated. The trigger is in the post process, eg. *.asnumpy().
So the inference latency would just be symbol constructing time not the execution time.

mms/metrics_manager.py
```diff
@@ -57,39 +72,64 @@ def start(metrics_write_to, namespace, mutex):
...
+                                                                  model_name=model_name,
+                                                                  unit=MetricUnit['ms'],
+                                                                  aggregate_method='interval_average',
+                                                                  write_to=metrics_write_to)
+
+        MetricsManager.metrics['host_cpu'] = Metric('host_cpu', mutex,
+                                                      model_name=None,
+                                                      unit=MetricUnit['percent'],
+                                                      is_resource=True,
```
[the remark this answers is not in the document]
yuruofeifei: Yes, I will change it.

mms/mxnet_model_server.py
```diff
@@ -156,15 +156,15 @@ def _arg_process(self):
             service_file = os.path.join(models[0][2], manifest['Model']['Service'])
 
             class_defs = self.serving_frontend.register_module(self.args.service or service_file)
-            
-            if len(class_defs) < 1:
-                raise Exception('User defined module must derive base ModelService.')
-            # The overrided class is the last one in class_defs
-            mode_class_name = class_defs[-1].__name__
+            class_defs = list(filter(lambda c: len(c.__subclasses__()) == 0, class_defs))
```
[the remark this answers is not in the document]
yuruofeifei: The check is already in model_service.py

mms/export_model.py
yuruofeifei: Mxnet models does not record the version when saving models.
And actually in our exporter, we are using onnx converter to convert onnx to mxnet. At here(https://github.com/awslabs/mxnet-model-server/blob/master/mms/export_model.py#L194), we are using local mxnet to serialize the model from memory to disk. So I think the version is still correct.

jesterhazy: 1. that path is only executed for onnx models, not for mxnet models
2. i think mxnet does record the version number. because i have seen errors like this at runtime:

src/nnvm/legacy_json_util.cc:185: Warning: loading symbol saved by MXNet version 1201 with lower version of MXNet v1200. May cause undefined behavior. Please update MXNet if you encounter any issue

yuruofeifei: Agree with what you said. 
Not all the mxnet model has version, at least the models I tried from model zoo does not have it. But I do find some model has version attr. 
For now I will add the version number from symbol if it has, otherwise the local mxnet version.

jesterhazy: 👍

mms/metric.py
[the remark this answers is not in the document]
lupesko: I'm good with that, as we discussed in person.

mms/metrics_manager.py
lupesko: I think any dahsboard tool will need the full statistical data to calculate percentiles, etc. So I do not think we're preventing integration with future dashboard providers.
I'm supportive of the usage of CW Statistics object.

mms/metrics_manager.py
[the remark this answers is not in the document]
lupesko: @yuruofeifei can you please file an issue for refactoring post-process, so we can have a proper post-process latency metric in the future?
We will handle this ticket post v0.2.

yuruofeifei:
Addressed all the comments.

yuruofeifei:
@jesterhazy ping for review

Thank you!

mms/metric.py
```diff
@@ -124,35 +152,56 @@ def start_recording(self):
...
+                    logger.info('Metric %s for last %s seconds is %f, writing to AWS CloudWatch...' %
+                        (self.name, self.interval_sec, metric))
+                    try:
+                        metric_data = {
+                            'MetricName': self.name,
+                            'Timestamp': utcnow,
+                            'Unit': self.unit,
+                            'StatisticValues': {
+                                'SampleCount': self.interval_datapoints_count,
+                                'Sum': self.interval_metric_aggregate,
```
jesterhazy: - how does this work for % values (i think it won't)
- we should report disk/mem used/free mb values as single data points not stats
- it looks like we will not be able to get percentiles this way, and would have to calculate and report those as a separate group of metrics.  i'm ok with deferring those to future enhancement.

yuruofeifei: 1. It will work for % values. The only percentage value we are collecting is cpu/memory/disk usage which is reported by python library. It is just a value of float number smaller 1, there should be no difference.
2. Changed.
3. Yea, we can do it later considering current timeline.

PR #298, opened 2018-01-23
Fix epoch number validation

knjcode:
In the current implementation, only alphanumeric characters and
underscore can be specified in the model name.

Currently, epoch number validation fails in case of model name
`imagenet1k-nin-0000.params`. In addition, validation also fails in
case `squeezenet_v1.1-0000.params` is shown in the export example.

So I changed the regular expression in epoch number validation from
`r'^\w+-\d+\.params$'` to `r'^.+-\d+\.params$'` so that symbols other
than underscore are allowed.

### DIFF
--- a/mms/export_model.py
+++ b/mms/export_model.py
@@ -240,7 +240,7 @@ def find_unique(files, suffix):
 
 
 def validate_epoch_number(params_file):
-    if re.match(r'^.+-\d+\.params$', params_file) is None:
+    if re.match(r'^[\w\-\.]+-\d+\.params$', params_file) is None:
         raise ValueError(NO_EPOCH_NUMBER_MESSAGE.format(params_file))
 
 
--- a/mms/export_model.py
+++ b/mms/export_model.py
@@ -240,7 +240,7 @@ def find_unique(files, suffix):
 
 
 def validate_epoch_number(params_file):
-    if re.match(r'^\w+-\d+\.params$', params_file) is None:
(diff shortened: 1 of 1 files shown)


mms/export_model.py
```diff
@@ -240,7 +240,7 @@ def find_unique(files, suffix):
 
 
 def validate_epoch_number(params_file):
-    if re.match(r'^\w+-\d+\.params$', params_file) is None:
+    if re.match(r'^.+-\d+\.params$', params_file) is None:
```
jesterhazy: thanks for submitting this!

matching any character seems to broad -- that could include spaces or other non-url compatible characters. can you change it to `r'^[\w\-]+-\d+\.params$'` ?

knjcode: Thanks for your comment.

How about using `r'^[\w\-\.]+-\d+\.params$'` so that `squeezenet_v1.1-0000.params` shown in the export example can also be used. (allow hyphens and dots.)

jesterhazy: yes, that would be perfect.

knjcode: Thanks.
Correct the code and commit it.

knjcode:
I modified the regex from `r'^.+-\d+\.params$'` to `r'^[\w\-\.]+-\d+\.params$'` to allow only alphanumeric characters, hyphens and dots.

lupesko:
Thank you @knjcode for contributing!

Issue #300, opened 2018-01-23
Race condition when multiple gunicorn worker download models with same name


yuruofeifei:
In our docker container, when multiple gunicorn worker download same models from S3 to disk, there will be read-write race condition because models are with the same name.

My proposal to fix it is adding a new optional argument: extract-to (Recommendation?) where the model will be saved and extracted to. 
Then we can create directory including gunicorn worker pid so models are separate across workers. It won't affect our current local serving flow but just requires some doc changes.
This argument would also be helpful if someone want to extract the model in different location rather than cwd.

yuruofeifei:
@jesterhazy @lupesko @aaronmarkham appreciate your feedback.

jesterhazy:
we need to fix this in a way that eliminates the duplicated work. 

given the current startup sequence, the most straightforward way to do this would to have only one of the model server workers do the download / extract part of the model loading.

we can use the fasteners to do this... specifically the \@fasters.interprocess_locked() decorator. 

factor the relevant part of _extract_model out into its own method, and wrap that method in the  \@fasters.interprocess_locked() decorator. add checks for model existence just outside and again inside the lock boundary (i.e. double-checked locking).

yuruofeifei:
I like this idea @jesterhazy , It won't affect any flow right now.

yuruofeifei:
Fixed via #302

jesterhazy:
fixed by #302

PR #302, opened 2018-01-23
Fix race condition when multiple gunicorn workers try to download same models

yuruofeifei:
Fix race condition when multiple gunicorn workers try to download same models. #300 
The fix is adding lock between processes to make sure only one worker is downloading and extracting.

Tested on a docker container, no error messages occurs and serving is successful.

@jesterhazy @lupesko

### DIFF
--- a/mms/model_loader.py
+++ b/mms/model_loader.py
@@ -15,6 +15,7 @@ import zipfile
 import shutil
 import json
 import jsonschema
+import fasteners
 
 from mms.log import get_logger
 from jsonschema import validate
@@ -26,14 +27,16 @@ URL_PREFIX = ('http://', 'https://', 's3://')
 MANIFEST_DIR = "manifest_schema"
 MANIFEST_SCHEMA_FILE = 'manifest-schema.json'
 MANIFEST_FILENAME = 'MANIFEST.json'
+LOCK_FILE = '/tmp/tmp_lock_file'
 
-def download(url, path=None, overwrite=False):
+@fasteners.interprocess_locked(LOCK_FILE)
+def _download_and_extract(model_location, path=None, overwrite=False):
     """Download an given URL
 
     Parameters
     ----------
-    url : str
-        URL to download
+    model_location : str
+        Location for local model or URL to download
     path : str, optional
         Destination path to store downloaded file. By default stores to the
         current directory with same name as in url.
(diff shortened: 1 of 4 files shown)


mms/model_loader.py
```diff
@@ -26,7 +27,20 @@
 MANIFEST_DIR = "manifest_schema"
 MANIFEST_SCHEMA_FILE = 'manifest-schema.json'
 MANIFEST_FILENAME = 'MANIFEST.json'
+LOCK_FILE = '/tmp/tmp_lock_file'
 
+@fasteners.interprocess_locked(LOCK_FILE)
```
jesterhazy: we should have one lock around complete download/extract process

mms/model_loader.py
```diff
@@ -26,7 +27,20 @@
...
+    if not os.path.isdir(model_dir):
+        os.mkdir(model_dir)
+    try:
+        if '.model' in model_file:
+            _extract_zip(model_file, model_dir)
+    except Exception as e:
+        raise Exception('Failed to open model file %s for model %s. Stacktrace: %s'
+                        % (model_file, model_file_prefix , e))
+
+@fasteners.interprocess_locked(LOCK_FILE)
```
jesterhazy: we are still going to download multiple times

mms/model_loader.py
```diff
@@ -26,7 +27,20 @@
...
 MANIFEST_SCHEMA_FILE = 'manifest-schema.json'
 MANIFEST_FILENAME = 'MANIFEST.json'
+LOCK_FILE = '/tmp/tmp_lock_file'
 
+@fasteners.interprocess_locked(LOCK_FILE)
+def _extract_model_archive(model_dir, model_file):
+    if not os.path.isdir(model_dir):
+        os.mkdir(model_dir)
+    try:
+        if '.model' in model_file:
```
jesterhazy: this check should happen before this method is even called. especially now that getting here requires lock acquisition

mms/model_loader.py
```diff
@@ -103,14 +117,8 @@ def _extract_model(service_name, path):
...
 
-    if not os.path.isdir(model_dir):
-        os.mkdir(model_dir)
-    try:
-        if '.model' in model_file:
-            _extract_zip(model_file, model_dir)
-    except Exception as e:
-        raise Exception('Failed to open model file %s for model %s. Stacktrace: %s'
-                        % (model_file, model_file_prefix , e))
+    _extract_model_archive(model_dir, model_file)
```
jesterhazy: single lock boundary should span everything from download to extract, inclusive

mms/model_loader.py
```diff
@@ -45,28 +48,45 @@ def download(url, path=None, overwrite=False):
...
+    model_file = model_location
+    if model_location.lower().startswith(URL_PREFIX):
+        if path is None:
+            model_file = model_location.split('/')[-1]
+        elif os.path.isdir(path):
+            model_file = os.path.join(path, model_location.split('/')[-1])
+        else:
+            model_file = path
+
+        if overwrite or not os.path.exists(model_file):
```
jesterhazy: each worker process will download and extract model to same location, duplicating work.

the lock will serialize access to this function, so results will be correct, but startup time will be multiplied by # of workers.

yuruofeifei: Yes, model will be extracted and download multiple times. But it should be according to the overwrite parameter. If we change the default behavior to always check existence, we are breaking the flow.

jesterhazy: ok we will fix that separately

Issue #312, opened 2018-01-26
update docker scripts


jesterhazy:
docker images do not build properly with onnx. fix.

PR #313, opened 2018-01-26
update docker scripts

jesterhazy:
*Issue #, if available:* #312 

*Description of changes:*

- update `install.sh` so image build runs correctly
- update gpu image so it has version of cuda required by mxnet-cu90

### DIFF
--- a/docker/Dockerfile.gpu
+++ b/docker/Dockerfile.gpu
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:8.0-cudnn5-devel
+FROM nvidia/cuda:9.0-cudnn7-devel
 
 COPY install.sh /
 
--- a/docker/install.sh
+++ b/docker/install.sh
@@ -1,18 +1,13 @@
 #!/usr/bin/env bash
 
 apt-get update
-apt-get install -y build-essential libatlas-base-dev libopencv-dev graphviz
-apt-get install -y python-setuptools python-pip
-apt-get install -y openjdk-7-jdk
-apt-get install -y nginx
-
 apt-get install -y software-properties-common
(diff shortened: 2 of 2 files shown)


docker/Dockerfile.gpu
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:8.0-cudnn5-devel
+FROM nvidia/cuda:9.0-cudnn7-devel
```
lupesko: Have we tested the GPU container?

docker/install.sh
```diff
@@ -1,18 +1,13 @@
...
-apt-get install protobuf-compiler libprotoc-dev
 
-pip install --upgrade pip
-pip install gunicorn
-pip install gevent
-pip install mxnet-model-server
+apt-get install -y build-essential libatlas-base-dev libopencv-dev graphviz python-dev \
+    openjdk-8-jdk nginx protobuf-compiler libprotoc-dev python-certbot-nginx curl
+
+cd /tmp && curl -O https://bootstrap.pypa.io/get-pip.py && python get-pip.py
```
lupesko: hmm. seems a bit unclean... why not something along the lines of:
```
$ apt-get install python-pip 
$ pip install --upgrade pip 
```

jesterhazy: apt-get repo may not have latest version of pip (usually doesn't). this is the best way to install pip.

docker/Dockerfile.gpu
jesterhazy: yes

docker/install.sh
lupesko: Got it. Cool.

PR #325, opened 2018-02-14
Change CPU docker image to install mxnet-mkl

charlieyou:


docker/Dockerfile.cpu
```diff
@@ -2,8 +2,8 @@ FROM ubuntu:latest
 
 COPY install.sh /
 
-RUN /install.sh
+RUN /install.sh && pip uninstall --yes mxnet && pip install mxnet-mkl
```
jesterhazy: last time i tried this, it installed a version of mkl build with 'experimental' flag set, and this let to correctness problems and memory leaks. have you tested this?

aaronmarkham: Just to muddy the waters...

Seems weird to run all of this stuff in a script and not how a typical Dockerfile is executed. Here, you have to look at two different files to see what's going on (install.sh and the dockerfile). 
Each apt-get install or other command can be on a RUN line. Then you can cleanup/optimize how docker handles the commands, taking advantage of caching and layers as needed. 
I found this [best practices guide helpful](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/).

Also, why not just install mxnet-mkl first then mms over the top? Won't MMS detect that MXNet is there and not touch it? 
Why install it via mms, just to uninstall it, and install it again as another flavor?

charlieyou: @jesterhazy Correctness seemed fine, was getting the same results, how did you test for memory leaks?

@aaronmarkham Pip just writes over the MXNet-MKL if it is installed first.
@lupesko Should refactoring of the docker files be passed on to whoever is working on containers?

lupesko: @jesterhazy @charlieyou I believe that MKL was fixed recently by intel.
@charlieyou yes, refactoring the Docker as proposed by @aaronmarkham is a great idea, and should be handled separately. Please file an issue to capture that and reference @vdantu

lupesko: @charlieyou let's make sure the issues @jesterhazy experienced were fixed in mxnet-mkl. Can you clarify how you tested for correctness before we merge this?

charlieyou: I sent some manual requests to each server just to make sure everything matched and also skimmed through the logs after each jmeter run. Everything was consistent as far as I could tell. Possible that if the error popped up 1 in 100 requests or so I would have missed it. Or if the incorrectness is in a component not used by resnet I also would have missed it.

lupesko: Thanks @charlieyou - LGTM.

Issue #282, opened 2018-01-16
folder cleanup after exporting an onnx model


aaronmarkham:
When you use mxnet-model-export on an onnx model, it creates some temp files: a param and symbol.json and a manifest file, before zipping those up in the model file. It should delete those temp files. It should not delete files that were already there.

PR #327, opened 2018-02-19
Removing temporary files created while exporting onnx models to mms

ankkhedia:
*Issue #, if available:*
Issue #282 
While importing onnx models for mms, some temporary files get created which are later bundled into .model files. These extra files get created and remains on the disk though it is not required later

*Description of changes:*
Instead of zipping all the files(existing+ temp) together, I zip temporary files as a separate step and delete them once they get zipped.

### DIFF
--- a/mms/export_model.py
+++ b/mms/export_model.py
@@ -276,46 +276,47 @@ def export_model(model_name, model_path, service_file=None, export_file=None):
     """
     Internal helper for the exporting model command line interface.
     """
-
-    if export_file is None:
-        export_file = '{}/{}.model'.format(os.getcwd(), model_name)
-    assert not os.path.exists(export_file), "model file {} already exists.".format(export_file)
-
-    if model_path.startswith('~'):
-        model_path = os.path.expanduser(model_path)
-
-    files = os.listdir(model_path)
     temp_files=[]
-    onnx_file = find_unique(files, '.onnx')
-    symbol_file = find_unique(files, '-symbol.json')
-    params_file = find_unique(files, '.params')
+    try:
+        if export_file is None:
+            export_file = '{}/{}.model'.format(os.getcwd(), model_name)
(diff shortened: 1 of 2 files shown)


ankkhedia:
@nswamy  The changes for cleaning up temp files created on model export.

mms/export_model.py
```diff
@@ -294,24 +294,26 @@ def export_model(model_name, model_path, service_file=None, export_file=None):
...
 
     manifest = generate_manifest(symbol_file, params_file, service_file, signature_file, model_name)
     with open(os.path.join(model_path, MANIFEST_FILE_NAME), 'w') as m:
         json.dump(manifest, m, indent=4)
-    files.append(MANIFEST_FILE_NAME)
+    temp_files.append(MANIFEST_FILE_NAME)
 
     with zipfile.ZipFile(export_file, 'w') as z:
         for f in files:
             z.write(os.path.join(model_path, f), f)
```
lupesko: Can we try to iterate over a combined list of files and temp_files, instead of duplicating the zip write statement?
We can then have a separate loop for removing the temp files.

mms/export_model.py
```diff
@@ -294,24 +294,26 @@ def export_model(model_name, model_path, service_file=None, export_file=None):
...
-
+        for f in temp_files:
+            z.write(os.path.join(model_path, f), f)
+            os.remove(os.path.join(model_path,f))
```
lupesko: What happens if the function errors and program terminates after the temp files have been created, but before they were removed? User will have temp files written.
Can we use try/finally block to make sure temp files are removed always (if they exist)?

ankkhedia:
@lupesko @vdantu @nswamy @jesterhazy @aaronmarkham 
 The changes requested by @lupesko  has been addressed.

PR #349, opened 2018-03-28
Changes for optimal configurations on EC2 instance

ankkhedia:
*Issue #, if available:*

*Description of changes:*
The changes are related to optimal configurations for EC2 instances for GPU and CPU

### DIFF
--- a/docker/Dockerfile.gpu
+++ b/docker/Dockerfile.gpu
@@ -21,8 +21,9 @@ RUN apt-get update && \
     && cd /tmp \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py \
-    && pip install gevent gunicorn mxnet-model-server \
-    && pip install --no-cache-dir mxnet-cu90 \
+    && pip install gevent gunicorn \
+    && pip install mxnet-model-server \
+    && pip install --no-cache-dir mxnet-cu90mkl \
     && mkdir /mxnet_model_server
 
 COPY wsgi.py setup_mms.py mxnet-model-server.sh /mxnet_model_server/
--- a/docker/README.md
+++ b/docker/README.md
@@ -69,7 +69,7 @@ We can optionally update the **nginx** section of the `mms_app_*.conf` file for 
 
 Note the `server_name` entry. You can update `localhost` to be your public hostname, IP address, or just use the default `localhost`. This depends on where you expect to utilize the Docker image. (Server Name can be updated at run-time. 
 This option is mentioned in steps to run.)
-
+ The CPU and GPU configs have been optimised for C5.2xlarge and p3.8xlarge respectively for Resnet-18 model. You can modify the parameters according to your use case.
 The **nginx** section will look like this:
 
 ```
@@ -228,7 +228,7 @@ Notes on a couple of the parameters:
 * **models** - the model used when setting up service. By default it uses resnet-18, chnage this argument to use customized model.
 * **worker-class** - the type of Gunicorn worker processes. We configure by default to gevent which is a type of async worker process. Options are [described in the Gunicorn docs](http://docs.gunicorn.org/en/stable/settings.html#worker-class).
 * **limit-request-line** - this is a security-related configuration that limits the [length of the request URI](http://docs.gunicorn.org/en/stable/settings.html#limit-request-line). It is useful preventing DDoS attacks.
-
+* **num-GPUs** - Number of GPUs available for use.
 ```
     [MMS arguments]
     --models
--- a/docker/mms_app_gpu.conf
+++ b/docker/mms_app_gpu.conf
@@ -20,13 +20,18 @@ optional
 --metrics-write-to
 optional
 
+--num-GPUs
+optional
+
 [Gunicorn Arguments]
 
 --bind
 unix:/tmp/mms_app.sock
 
+###Number of Gunicorn workers optimised for p3.8xlarge instance ##########
 --workers
-1
+32
+
 
 --worker-class
 gevent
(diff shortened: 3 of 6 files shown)


ankkhedia:
@lupesko @jesterhazy @vdantu @aaronmarkham 
@awslabs/aws-deep-learning-sdk 
Could you please review the changes for optimal configurations for MMS on CPU and GPU ec2 instance.

docker/README.md
```diff
@@ -94,6 +94,10 @@ We can optionally update the **nginx** section of `mms_app_cpu.conf` or `mms_app
 * For CPU builds, use [mms_app_cpu.conf](mms_app_cpu.conf) and [Dockerfile.cpu](Dockerfile.cpu).
 * For GPU builds, use [mms_app_gpu.conf](mms_app_gpu.conf) and [Dockerfile.gpu](Dockerfile.gpu).
 
+
+
+ The CPU and GPU configs have been optimised for C5.2xlarge and p3.8xlarge respectively for Resnet-18 model. You can modify the parameters according to your use case.
```
vdantu: Its better to elaborate the parameters that need to be modified.

docker/README.md
```diff
@@ -263,8 +267,11 @@ Notes on a couple of the parameters:
 * **models** - the model used when setting up service. By default it uses resnet-18, chnage this argument to use customized model.
 * **worker-class** - the type of Gunicorn worker processes. We configure by default to gevent which is a type of async worker process. Options are [described in the Gunicorn docs](http://docs.gunicorn.org/en/stable/settings.html#worker-class).
 * **limit-request-line** - this is a security-related configuration that limits the [length of the request URI](http://docs.gunicorn.org/en/stable/settings.html#limit-request-line). It is useful preventing DDoS attacks.
+* **num-GPUs** - Number of GPUs available for use.
```
vdantu: It's better to specify that we will use [0..num-GPUs] for MMS.

docker/wsgi.py
```diff
@@ -13,15 +13,26 @@
...
 is_gpu_image = os.environ['MXNET_MODEL_SERVER_GPU_IMAGE']
-
-
+num_gpu = 1
+LOCK_FILE = '/tmp/tmp_lock_file'
+@fasteners.interprocess_locked(LOCK_FILE)
+def get_gpu_ids(args, n_gpu):
+    args.append('--gpu')
+    gpu_id=0
+    n_gpu =4
```
vdantu: hard coded to 4?

docker/wsgi.py
```diff
@@ -13,15 +13,26 @@
...
+    with open('/mxnet_model_server/.gpu_id', 'r') as file:
+        gpu_id = file.read()
```
vdantu: Are we assuming that the file will exist?

docker/wsgi.py
```diff
@@ -13,15 +13,26 @@
...
+        args.append(str(int(gpu_id )% n_gpu))
+    with open('/mxnet_model_server/.gpu_id', 'w') as file:
+        file.write(str(int(gpu_id) + 1))
```
vdantu: Why open the same file twice? Why not open it with "r+"

docker/wsgi.py
```diff
@@ -50,20 +61,21 @@ def read_models_from_file():
                 if line.startswith('['):
                     found_mms_args = 1 if mms_arg_header.lower() in line.lower() else 0
                 if found_mms_args is 1:
-                    if line.startswith('--') and content[i+1] != 'optional':
-                        args.append(line)
-                        if line.startswith('--model'):
-                            args += content[i + 1].split(' ')
-                        else:
-                            args.append(content[i + 1])
+                    if line.startswith('--num-GPUs') and content[i+1] != 'optional':
```
vdantu: option should be small-case.--num-gpu

docker/wsgi.py
```diff
@@ -50,20 +61,21 @@ def read_models_from_file():
...
+                        num_gpu = int(content[i+1])
```
vdantu: So if the num-gpus is given with a non-optional value, we wouldn't be reading models? They seem to be in a if/else statement. Also, default the "num_gpu" variable to "1" or some value here and use that in the "get_gpu_ids" def. Don't assign "num_gpu" inside the "get_gpu_ids" method.

docker/mms_app_gpu.conf
```diff
@@ -20,13 +20,18 @@ optional
 --metrics-write-to
 optional
 
+--num-GPUs
+optional
```
vdantu: What would we default to when this is optional? For now, I think we can hard code it to 1 to avoid  last minute churn.

docker/mxnet-model-server.sh
```diff
@@ -99,6 +99,7 @@ start_mms()
     service nginx restart
     # Download and extract all the models
     python /mxnet_model_server/setup_mms.py $models
+    echo "0" >& /mxnet_model_server/.gpu_id
```
vdantu: I would recommend moving this line above setup_mms.py or moving it inside setup_mms.py. The guinicorn instances (created below) shouldn't be created unless the setup and this echo go through.

docker/Dockerfile.gpu
```diff
@@ -23,7 +23,7 @@ RUN apt-get update && \
     && python get-pip.py \
     && pip install gevent gunicorn \
     && pip install --pre mxnet-model-server \
-    && pip install --no-cache-dir mxnet-cu90 \
+    && pip install --no-cache-dir mxnet-cu90mkl \
```
vdantu: +1

docker/wsgi.py
ankkhedia: r+ appends the file and does not overwrite

docker/wsgi.py
ankkhedia: The models need to be read only in case where the line starts with --model. Here the  line starts with -- num-gpus, so we don't need to read model for that line

Issue #350, opened 2018-03-30
Explain the parameters needed for fine-tuning performance


lupesko:
[This section](https://github.com/awslabs/mxnet-model-server/blob/master/docker/README.md#advanced-settings) details the config params.
It needs to explicitly call out how to tune parameters based on the runtime hardware available, and explain why/how we suggest updating the number of workers.

ankkhedia:
@lupesko I have added the docs about detailed configurations

lupesko:
Thanks @vdantu - closing.

PR #355, opened 2018-04-02
Latency explanations

ankkhedia:
*Issue #, if available:*
#350 

*Description of changes:*
Adds a note for explaining latency number differences in GPU and CPU


docs/optimised_config.md
```diff
@@ -40,6 +40,8 @@ We performed similar experiments for CPU (100 requests from 100 concurrent worke
 We also found similar results for CPU as shown in the above plots i.e throughput is highest when there are 8 workers(which equals number of vCPUs in c5.2xlarge)
 **Based on the results, we recommend setting number of Gunicorn workers equal to number of vCPUs present in the instance.**
 
+**Note:** The higher latencies seen during the load tests on CPU instance is because the c5.2xlarge CPU instance has 1/4th the total number of vCPUs as compared the p3.8xlarge GPU instance and both were serving the same number of incoming request. Apart from it, the GPU also shares some workload of CPUs avoiding any backlogs and hence better latency numbers. We had seen significantly lesser latencies when the rate of requests coming into c5.2xlarge instance were reduced by 1/4th.
```
vdantu: The higher latencies were seen during the load tests on CPU instance was because the c5.2xlarge CPU instance has 1/4th the total number of vCPUs as compared to the p3.8xlarge GPU instance and both were serving the same number of incoming request. Apart from it, the GPU also shares some workload of CPUs reducing the request backlogs. We had seen significantly lesser latencies when the rate of requests coming into c5.2xlarge instance was reduced by 1/4th

Issue #362, opened 2018-04-18
Update the mxnet pip package


vdantu:
There might be a need for two pip packages
1. CPU pip package which installs mxnet-mkl
2. GPU pip package which installs mxnet-cu90mkl

vrakesh:
Researched on this

Two possible approaches
1. using PIP --install-option flag - This approach currently fails at PIP level and works at setup.py level. Investigating the reason for failure.

2. Using nvidia-smi via sub-process - This approach will work . But will have to adapt to change in behavior of the tool over time , does it work even if GPU is removed physically, is a question we need to answer.

Right now the manual work around for this issue, while using gpu acceleration, one needs to

`pip uninstall mxnet-mkl`
`pip install mxnet-cu90mkl'

lupesko:
Can't we just use mxnet-cu90mkl since it supports both CPU and GPU?

vrakesh:
Tested out the mxnet-cu90mkl package , it works under Linux for both cpu and gpu.
But Mac and Windows are not supported by the cu90 package as of now

vrakesh:
fixed in pull request #366

PR #366, opened 2018-04-25
updating pip package to support mxnet-mkl or cu90mkl by default

vrakesh:
*Issue #, if available:*
#362 
*Description of changes:*

Now setup.py installs mxnet-mkl for non-linux distributions and mxnet-cu90mkl for linux distributions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,11 +23,13 @@ pkgs.append('tools')
 # $ pip install wheel
 # $ python setup.py bdist_wheel --universal
 # $ twine upload dist/*
-requirements = ['mxnet-mkl>=1.1', 'Flask', 'Pillow', 'requests', 'flask-cors',
+requirements = ['Flask', 'Pillow', 'requests', 'flask-cors',
                 'psutil', 'jsonschema', 'onnx-mxnet>=0.4.2', 'boto3', 'importlib2',
                 'fasteners']
 if platform.system() == 'Linux':
-    requirements[0] = 'mxnet-cu90mkl>=1.1'
+    requirements = ['mxnet-cu90mkl>=1.1'] + requirements
+else:
+    requirements = ['mxnet-mkl>=1.1'] + requirements
 setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -22,16 +23,19 @@
 # $ pip install wheel
 # $ python setup.py bdist_wheel --universal
 # $ twine upload dist/*
-
+requirements = ['mxnet-mkl>=1.1', 'Flask', 'Pillow', 'requests', 'flask-cors',
+                'psutil', 'jsonschema', 'onnx-mxnet>=0.4.2', 'boto3', 'importlib2',
+                'fasteners']
+if platform.system() == 'Linux':
+    requirements[0] = 'mxnet-cu90mkl>=1.1'
```
lupesko: This will work, but will be very easy to break in the future. E.g. someone adds a new dependency, and puts it in the first slot in the requirements vector...
Maybe instead we can do a if/else statement that inserts the right package (cpu/gpu) in the 0-th index?

vrakesh: agreed , will improve this

setup.py
```diff
@@ -22,16 +23,21 @@
 # $ pip install wheel
 # $ python setup.py bdist_wheel --universal
 # $ twine upload dist/*
-
+requirements = ['Flask', 'Pillow', 'requests', 'flask-cors',
+                'psutil', 'jsonschema', 'onnx-mxnet>=0.4.2', 'boto3', 'importlib2',
+                'fasteners']
+if platform.system() == 'Linux':
```
vdantu: Could you describe how you tested these changes? And on what systems.

vrakesh: Tested the changes the following machines
1. Ubuntu server ec2 instance
2. AMI ec2 instance.
3. PureOS(Debian) system.
4. MacOS laptop
The installation works on all these machines.

However , due to ongoing ONNX issue, it is not working DLAMI.

Issue #337, opened 2018-03-06
Do not allow multiple runs of MMS on a single instance


vdantu:
To manage the host-resources and keep the MMS instances updated with latest models set, there is a need to prevent multiple runs of MMS command. 

The suggestion is that if there exists a previous run of this command running on the host, avoid starting another instance of MMS on the same host. Instead throw an error back saying "There already exists one instance of running MMS"

vdantu:
This has been addressed in the latest version of MMS.

vrakesh:
Resolved in pull request #367

PR #367, opened 2018-04-26
Process duplicate

vrakesh:
*Issue #, if available:*
#337 
*Description of changes:*

Using pgrep to check if the process already exists, if it does we will not allow for the new process to start, recommends user to use stop or restart.

While running start command we check if a process is already running

### DIFF
--- a/docker/mxnet-model-server
+++ b/docker/mxnet-model-server
@@ -36,9 +36,19 @@ start_mms()
     # Check if the process is already running , if so let the user know and exit
     for pid in $(pgrep -f mxnet-model-server); do
         if [ $pid != $$ ]; then
-            echo "[$(date)] : mxnet-model-server : Process is already running with PID $pid, run mxnet-model-server stop or mxnet-model-server restart"
+            echo "[$(date)] : mxnet-model-server : Process is already running with PID $pid, run mxnet-model-server stop or mxnet-model-server restart --mms-config=/path/to/config"
             exit 1
         else
+          # Look for orphaned gunicorn workers, kill them if found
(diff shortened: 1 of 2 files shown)


docker/mxnet-model-server
```diff
@@ -18,7 +18,7 @@ NGINX_CONFIG_FILE='/etc/nginx/conf.d/virtual.conf'
 MMS_CONFIG_FILE=''
 
 # Start MxNet model server. This function expects one argument which is the application config file.
-start_mms() 
+start_mms()
 {
     gunicorn_arguments=''
     # TODO issue:https://github.com/awslabs/mxnet-model-server/issues/337. Check if MMS is already running.
```
lupesko: You are fixing this issue, so can remove the comment

docker/mxnet-model-server
```diff
@@ -34,6 +33,16 @@ start_mms()
     models=""
     models_found=0
 
+    # Check if the process is already running , if so let the user know and exit
+    for pid in $(pgrep -f mxnet-model-server); do
```
vdantu: Looks good. 
Do the gunicorn workers clean up if mxnet-model-server is killed?

docker/mxnet-model-server
```diff
@@ -34,6 +33,16 @@ start_mms()
...
+        if [ $pid != $$ ]; then
+            echo "[$(date)] : mxnet-model-server : Process is already running with PID $pid, run mxnet-model-server stop or mxnet-model-server restart"
```
vdantu: Good debug to have. Just finish up the restart command completely. I don't think restart can work without --mms-config option. Customers can just copy paste command from this.

Issue #364, opened 2018-04-18
CI for MMS performance


ankkhedia:
We need to benchmark MMS for CI to keep a check on performance regressions

ankkhedia:
Already merged the changes

PR #371, opened 2018-04-27
benchmark ai integration script

ankkhedia:
*Issue #, if available:*
#364 
CI for MMS

*Description of changes:*
This is the script which will be run by benchmark AI for MMS CI

### DIFF
--- a/ci/benchmark_ai.sh
+++ b/ci/benchmark_ai.sh
@@ -1,39 +1,36 @@
 #!/bin/bash
 
-#Get option to run test on GPU/CPU image
-while getopts "g:" option; do
-  case $option in
-    g)      use_gpu=$OPTARG;;
-  esac
-done
-echo $use_gpu
-
+##global variables
+use_gpu="0"
 report=report.csv
-
-echo "$report"
-
+num_workers=(50 100)
+num_requests=(25 50 100)
 
 # Install Docker CLI in case it is missing
 install_docker(){
     docker run --name hw hello-world >& /dev/null
     if [[ `echo $?` != 0 ]] ;then
-       echo "Something went wrong with docker CLI installation"
+       echo "Installing docker cli"
        # Install DOCKER CLI
-       apt-get remove docker docker-engine docker.io
-       apt-get update
+       apt-get remove docker docker-engine docker.io &&
+       apt-get update &&
        apt-get install -y \
             apt-transport-https \
             ca-certificates \
             curl \
-            software-properties-common
-       curl -fsSL https://download.docker.com/linux/ubuntu/gpg |  apt-key add -
-       apt-key fingerprint 0EBFCD88
+            software-properties-common && 
+       curl -fsSL https://download.docker.com/linux/ubuntu/gpg |  apt-key add - &&
+       apt-key fingerprint 0EBFCD88 &&
        add-apt-repository \
 	"deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-       apt-get update
+       apt-get update &&
        apt-get install  -y docker-ce=18.03.1~ce-0~ubuntu
+       if [[ `echo $?` != 0 ]] ; then
+           echo "ERROR: Exiting as docker cli installation failed"
+           exit 1
+       fi
     else
-        echo "Docker is happy"
+        echo "Docker already installed"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


ci/benchmark_ai.sh
```diff
@@ -0,0 +1,149 @@
...
+
+#start the container test
+for workers in 50 100
+do
+    for requests in 25 50 100
+    do
+        echo "*************************************** Workers $workers and reqs $requests   ****************************************************"
+        if [ "$use_gpu" == "0" ]; then
+            docker run  --name mms -v /mxnet-model-server:/mxnet-model-server  -itd -p 80:8080 awsdeeplearningteam/mms_cpu
+            docker exec mms pip install -U -e /mxnet-model-server/.
```
vdantu: Can this if/else be abstracted out to a seperate function?

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,149 @@
...
+            nvidia-docker exec mms pip install -U -e /mxnet-model-server/.
+            nvidia-docker exec mms pip uninstall --yes mxnet-cu90mkl
+            nvidia-docker exec mms pip uninstall --yes mxnet
+            nvidia-docker exec mms pip install mxnet-cu90mkl
+            nvidia-docker exec mms mxnet-model-server start --mms-config /mxnet-model-server/docker/mms_app_gpu.conf >& /dev/null &
+            sleep 90
+        fi
+
+        cd /mxnet-model-server/load-test
+        export JMETER_HOME=$HOME/apache-jmeter-4.0
```
vdantu: Do these exports change for every run of the test? If not, we should do this at the beginning of the test

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,149 @@
...
+        export PATH=$JMETER_HOME/bin:$PATH
+        ./run_load_test.sh -i 127.0.0.1 -c $workers -n $requests -f report.csv -p 80
+        echo "done report generation"
+        OLDIFS= $IFS
+        IFS=","
+        read -r sampler_label aggregate_report_count average aggregate_report_median aggregate_report_90_line aggregate_report_95_line aggregate_report_99_line aggregate_report_min aggregate_report_max a\
```
vdantu: Move this parsing of log file to a seperate function, maybe parse_and_publish_metrics(). Take report.csv as a global variable, so that it can be referenced in the parse_and_publish_metrics()

ankkhedia:
@vdantu 
I have addressed all the comments. Could you please review it ?

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
+#!/bin/bash
+
+#Get option to run test on GPU/CPU image
+while getopts "g:" option; do
```
vdantu: Just curious, why not do this processing in main?

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+  case $option in
+    g)      use_gpu=$OPTARG;;
+  esac
+done
+echo $use_gpu
+
+report=report.csv
```
vdantu: Its better to move globals to the top. That way we don't mistake them for a command.

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+
+echo "$report"
+
```
vdantu: Extra space

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+        apt-get update
+        apt-get install -y nvidia-docker2
+        pkill -SIGHUP dockerd
+    else
+        echo "nvidia docker already installed"
+    fi
+
+}
+
+
```
vdantu: Extra space

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+}
+
+#JMETER installation
+install_jmeter(){
+    echo "jmeter installation"
+    if [ -d "$HOME/apache-jmeter-4.0" ]; then
+        echo "JMeter already installed"
+    else
+        echo "Installing jmeter"
+        cd /tmp
```
vdantu: Should we append "&&" to the end of each command so that if one of the commands fails, the main function will get that error?? If one of the commands fails here, we will proceed with the rest of the commands right?

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+            parse_csv_log
+            docker rm -f mms
+        done
+    done
+}
+
+
+main(){
+    install_docker
+    install_jmeter
```
vdantu: If we add "&&" to each of the running commands in the above methods, we should probably check for return values ("echo $?") after every step here.

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+        install_nvidia_docker
+    fi
+    prepare_repo
+    pull_docker_images
+    run_test
+    docker rm -f hw
+}
+
+main
+exit 0
```
vdantu: Wouldn't this always return a success no matter what happens with main run? You might want to consider removing the "exit 0"

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+}
+
+remove_container_images(){
+    docker rm -f mms
+}
+
+run_test(){
+    export JMETER_HOME=$HOME/apache-jmeter-4.0
+    export PATH=$JMETER_HOME/bin:$PATH
+    for workers in 50 100
```
vdantu: This also seems like hard-coded value. You could take this as variable array at the top of the file as follows
"num_workers=(50 100)"
and you can use num_workers here
 "${num_workers[@]}"

Not a blocker for commit since this is the only place you use this values.

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,218 @@
...
+    do
+        for requests in 25 50 100
```
vdantu: Similarly here

ankkhedia:
@vdantu I think I addressed your remaining comments. Could you please review it?

ci/benchmark_ai.sh
```diff
@@ -0,0 +1,246 @@
+#!/bin/bash
+
+##global variables
```
vdantu: Cool. Thanks for moving these up

PR #375, opened 2018-05-08
Batching: DataStore + related components

charlieyou:
Part of the Batching MVP Implementation

DataStore is the API that is used for communication between the ModelThread and PredictFrontend when batching is enabled.
It abstracts away the Queue, HashMap, and Codecs. The Queue and HashMap have implementations for their associated Redis data structures.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #377, opened 2018-05-09
Batching: DataStore + PredictFrontend

charlieyou:
Same as #375, but adding PredictFrontend to get CR comments on it.

PredictFrontend abstracts the predict callback so that Redis is bypassed when batching is not used.


mms/predict_frontend.py
```diff
@@ -0,0 +1,165 @@
...
+                logger.error(str(e))
+                abort(400, str(e))
+        else:
+            msg = '%s is not supported for input content-type' % input_type
+            if model_name + '_Prediction5XX' in MetricsManager.metrics:
+                MetricsManager.metrics[model_name + '_Prediction5XX'].update(metric=1)
+            logger.error(msg)
+            abort(500, "Service setting error. %s" % (msg))
+
+        if self.batching:
```
vdantu: Shouldn't this be in try/except block?

mms/predict_frontend.py
```diff
@@ -0,0 +1,165 @@
...
+                MetricsManager.metrics[model_name + '_Prediction5XX'].update(metric=1)
+            logger.error(msg)
+            abort(500, "Service setting error. %s" % msg)
+
+        logger.debug("Prediction request handling time is: %s ms",
+                     ((time.time() - handler_start_time) * 1000))
+        return self.handler.jsonify({'prediction': response})
+
+    def _batch_predict(self, input_data, input_type, output_type, model_name):
+        _id = str(uuid.uuid4())
```
vdantu: Does this ensure uniqueness? The responses from model-services are mapped to a particular frontend thread based on this right?

mms/serving_frontend.py
```diff
@@ -40,10 +40,23 @@ def __init__(self, app_name):
...
+        def gen_prefix(prefix, length=5, chars=string.ascii_letters+string.digits):
+            return prefix + ''.join(random.choice(chars) for _ in range(length)) + '_'
+
         try:
             self.service_manager = ServiceManager()
             self.handler = FlaskRequestHandler(app_name)
 
+            if batching:
+                # generate a pseudo-random prefix to avoid collisions between runs
+                prefix = gen_prefix(app_name)
```
vdantu: Wouldn't this tie each front-end to its own model-services? You might want to consider removing this if you want other model-service threads to handle the queues.

docker/redis.conf
```diff
@@ -0,0 +1,47 @@
...
+
+################################# GENERAL #####################################
+daemonize yes
+supervised no
+pidfile redis.pid
+loglevel notice
+logfile redis.log
+
+################################ SNAPSHOTTING  ################################
+save 900 1
```
vdantu: Based on our discussion, disable this snapshotting for now. If the requirements exists to snapshot requests, we can re-do the changes.

mms/serving_frontend.py
charlieyou: fixed

mms/predict_frontend.py
charlieyou: Yes, here is a SO answer on uuid4: https://stackoverflow.com/questions/703035/when-are-you-truly-forced-to-use-uuid-as-part-of-the-design/786541#786541

"Frankly, in a single application space without malicious actors, the extinction of all life on earth will occur long before you have a collision, even on a version 4 UUID, even if you're generating quite a few UUIDs per second."

mms/predict_frontend.py
charlieyou: done

mms/serving_frontend.py
```diff
@@ -40,17 +38,14 @@ def __init__(self, app_name, batching=False, data_store_config=None):
...
-
         try:
             self.service_manager = ServiceManager()
             self.handler = FlaskRequestHandler(app_name)
 
             if batching:
-                # generate a pseudo-random prefix to avoid collisions between runs
-                prefix = gen_prefix(app_name)
-                data_store = DataStore(prefix, data_store_config)
+                # TODO generate a unique prefix in docker/mxnet-model-server and pass it into all gunicorn workers
```
vdantu: Good point to make a TODO. would this still be an issue after removing snapshotting?

Issue #340, opened 2018-03-08
Generic integration test framework


ankkhedia:
The integration tests right now supports only image based models and hence uses hardcoded kitten image as input. The tests should be general so that it can address all the models in the model zoo. Also, all the references should be put to mms S3 bucket to avoid failures in case of any change to onnx buckets .

This can be done by pulling models from S3 bucket with input, expected output and other artefacts about models kept in same folder.  Also make sure that all the outputs are correct rather than just checking if we get a results

PR #419, opened 2018-05-31
Adding FERPlus model to model zoo

lupesko:
*Description of changes:*
Adding FERPlus model (emotion detection CNN) to model zoo, this model is imported from ONNX

@aaronmarkham @ankkhedia

### DIFF
--- a/mms/tests/integration_tests/test_utils.py
+++ b/mms/tests/integration_tests/test_utils.py
@@ -31,19 +31,20 @@ onnx_mxnet_model_urls = {
     'onnx-inception_v1': 'https://s3.amazonaws.com/model-server/models/onnx-inception_v1/inception_v1.onnx',
     'onnx-vgg19': 'https://s3.amazonaws.com/model-server/models/onnx-vgg19/vgg19.onnx'}
 
-# models from onnx-mxnet model zoo
+# models from mxnet-model-server model zoo
 mxnet_model_urls = {
-    # TODO: check why these two models are failing github issue #345
     'caffenet': 'https://s3.amazonaws.com/model-server/models/caffenet/caffenet.model',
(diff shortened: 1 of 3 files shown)


ankkhedia:
@lupesko The integration test was written when we had only CNN models trained on imagenet and it is not generic enough for all types of inputs/outputs. 
There is an open issue for a generic framework which needs to be taken up. 
Issue #340

mms/tests/integration_tests/test_utils.py
```diff
@@ -44,6 +44,7 @@
     'resnext-101-64x4d': 'https://s3.amazonaws.com/model-server/models/resnext-101-64x4d/resnext-101-64x4d.model',
     'squeezenet_v1.1': 'https://s3.amazonaws.com/model-server/models/squeezenet_v1.1/squeezenet_v1.1.model',
     'vgg19': 'https://s3.amazonaws.com/model-server/models/vgg19/vgg19.model',
+    'ferplus': 'https://s3.amazonaws.com/model-server/models/FERPlus/FERPlus.model',
```
aaronmarkham: This won't work. If the data input is `Input2505` for the input layer, the test utils only expects to see `data` or `input0`. I noticed this with K's keras blog - it used `/input11` or something. I thought the slash was a problem and the input layer deviation was super interesting. Really need to think through the implications of this for the zoo, mms, and how stuff is tested.

lupesko: Yes, note my comments above. The integration suite code needs to be refactored to be able and handle broader set of models.

lupesko:
@aaronmarkham @ankkhedia updated the PR to include only the added model.
Will add test improvements later as part of #340 
Please review.

aaronmarkham:
Lgtm

ankkhedia:
PR looks good!

PR #433, opened 2018-07-06
Add license text to java file.

frankfliu:
*Issue #, if available:*

*Description of changes:*

1. Add license text to all java file.
2. Allow start local MMS with gradle.


frontend/tools/gradle/launcher.gradle
```diff
@@ -0,0 +1,109 @@
+apply plugin: LauncherPlugin
```
vdantu: Could we add license to this as well?

frontend/README.md
```diff
@@ -16,12 +16,19 @@ You will find a jar file in frontend/server/build/libs file.
...
-Frontend web server using a configuration file to controll the behavior of the frontend web server.
+Frontend web server using a configuration file to control the behavior of the frontend web server.
 An sample config.properties can be found in frontend/server/src/test/resources/config.properties.
 This configure will load a noop model by default. The noop model file is located in frontend/modelarchive/src/test/resources/model/noop-v0.1.model.
 
+#### Start Query service:
+
+```sh
+cd frontend/server
+../gradlew startServer
```
vdantu: What prevents us from using this from frontend directly? Why use "../". Also should we move gradlew to another directory like "mxnet-model-server/tools" or create a new "bin" directory in MMS? We could export that as a path variable and use it to start the server in containers.

frankfliu: I added a symbolic link to frontend/server folder.
All gradlew files should be in the root project directory. If user installed gradle locally or they added gradlew to the PATH, then they don't have to use "../".

"startServer" is task defined in server module, that's why you need to execute it in server folder. You can also execute in root project folder, but you have to specify which module you what to use by "gradlew -P server startServer", this is the same convention for all the build system that looking for build file in current directory.

PR #436, opened 2018-07-07
Implement new REST API

frankfliu:
Implement new REST API.
Implement model archive 1.0 format.


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -65,8 +91,8 @@ public boolean unregisterModel(String modelName) throws WorkerInitializationExce
             return false;
         }
 
-        model.setMinWorker(0);
-        model.setMaxWorker(0);
+        model.setMinWorkers(0);
+        model.setMaxWorkers(0);
```
vdantu: We should remove these two parameters and change it to "numWorkers". I don't see a point in having two values when we support only scaling to a particular level.

mms/log.py
```diff
@@ -99,6 +85,6 @@ def get_logger(name=None, level="NOTSET"):
     return logger
 
 
-def log_msg(*args, **kwargs):
+def log_msg(*args):
```
vdantu: Good catch. It might be needed though, if we are passing a set of key-value pairs. I will work on added such a log capability.

mms/model_service_worker.py
```diff
@@ -261,7 +237,7 @@ def predict(self, data):
             batch_size = len(req_batch)  # num-inputs gives the batch size
             input_batch, req_id_map, invalid_reqs = self.retrieve_data_for_inference(req_batch, model_service)
             if batch_size == 1:
-                retval.append(model_service.inference([input_batch[0][i] for i in input_batch[0]]))
+                retval.append(model_service.inference(input_batch[0]))
```
vdantu: This should probably be reverted back. How would we pass multiple inputs to a single request?

mms/utils/validators/validate_model_artifacts.py
```diff
@@ -1,46 +0,0 @@
-# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-# Licensed under the Apache License, Version 2.0 (the "License").
```
vdantu: Why are we removing this file?

PR #451, opened 2018-07-15
1. Added a per thread queue for Jobs.

vdantu:
*Issue #, if available:*

*Description of changes:*
1. Monitor synchronous messages to the backend worker.
1. Restart the backend worker if there is no response in 2 mins.
1. Added a thread based queue to send control messages to specific backend workers.
1. Refactored code a little.

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/BaseModelRequest.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/BaseModelRequest.java
@@ -14,15 +14,15 @@ package com.amazonaws.ml.mms.util.messages;
 
 public class BaseModelRequest {
 
-    private String command;
+    private WorkerCommands command;
     private String modelName;
 
-    public BaseModelRequest(String command, String modelName) {
+    public BaseModelRequest(WorkerCommands command, String modelName) {
         this.command = command;
         this.modelName = modelName;
     }
 
-    public String getCommand() {
+    public WorkerCommands getCommand() {
         return command;
     }
 
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/ModelLoadModelRequest.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/ModelLoadModelRequest.java
@@ -27,11 +27,11 @@ public class ModelLoadModelRequest extends BaseModelRequest {
     private String gpu;
 
     public ModelLoadModelRequest() {
-        super("load", null);
+        super(WorkerCommands.LOAD, null);
     }
 
     public ModelLoadModelRequest(Model model, String gpu) {
-        super("load", model.getModelName());
+        super(WorkerCommands.LOAD, model.getModelName());
         this.handler = model.getModelArchive().getManifest().getModel().getHandler();
         this.batchSize = model.getBatchSize();
         this.gpu = gpu;
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
@@ -19,10 +19,8 @@ import com.amazonaws.ml.mms.util.messages.ModelWorkerResponse;
 import com.amazonaws.ml.mms.util.messages.Predictions;
 import com.amazonaws.ml.mms.util.messages.RequestBatch;
 import java.nio.charset.Charset;
-import java.util.ArrayList;
 import java.util.Base64;
 import java.util.LinkedHashMap;
-import java.util.ListIterator;
 import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -39,11 +37,11 @@ public class BatchAggregator {
         jobs = new LinkedHashMap<>();
     }
 
-    public BaseModelRequest getRequest() throws InterruptedException {
+    public BaseModelRequest getRequest(String threadName) throws InterruptedException {
         jobs.clear();
 
         // first job is a blocking call;
-        Job job = model.nextJob();
+        Job job = model.nextJob(threadName);
         if (job.isControlCmd()) {
             RequestBatch input = job.getPayload();
             String gpu = input.getStringParameter("gpu");
@@ -58,21 +56,7 @@ public class BatchAggregator {
         int size = model.getBatchSize() - 1;
         long begin = System.currentTimeMillis();
         for (int i = 0; i < size; ++i) {
-            if (job.isControlCmd()) {
-                jobs.remove(job.getJobId());
-
-                ListIterator<Job> iterator =
-                        new ArrayList<>(jobs.values()).listIterator(jobs.size());
-                while (iterator.hasNext()) {
-                    Job j = iterator.next();
-                    model.addFirst(j);
-                }
-                RequestBatch input = job.getPayload();
-                String gpu = input.getStringParameter("gpu");
-                return new ModelLoadModelRequest(model, gpu);
-            }
-
-            job = model.nextJob(maxBatchDelay);
+            job = model.nextJob(threadName, maxBatchDelay);
             if (job == null) {
                 break;
             }
@@ -141,18 +125,34 @@ public class BatchAggregator {
             return;
         }
 
-        ModelInferenceRequest msg = (ModelInferenceRequest) message;
-        for (RequestBatch req : msg.getRequestBatch()) {
-            String requestId = req.getRequestId();
-            Job job = jobs.remove(requestId);
-            if (job == null) {
-                throw new IllegalStateException("Unexpected job: " + requestId);
+        if (message != null) {
+            ModelInferenceRequest msg = (ModelInferenceRequest) message;
+            for (RequestBatch req : msg.getRequestBatch()) {
+                String requestId = req.getRequestId();
+                Job job = jobs.remove(requestId);
+                if (job == null) {
+                    throw new IllegalStateException("Unexpected job: " + requestId);
+                }
+                job.sendError(error);
             }
-            job.sendError(error);
-        }
+            if (!jobs.isEmpty()) {
+                jobs.clear();
+                throw new IllegalStateException("Not all jobs get response.");
+            }
+        } else {
+            // Send the error message to all the jobs
+            for (Map.Entry<String, Job> j : jobs.entrySet()) {
+                String jobsId = j.getValue().getJobId();
+                Job job = jobs.remove(jobsId);
 
-        if (!jobs.isEmpty()) {
-            throw new IllegalStateException("Not all jobs get response.");
+                if (job.isControlCmd()) {
+                    job.sendError(error);
+                } else {
+                    // Data message can be handled by other workers.
+                    // If batch has gone past its batch max delay timer?
+                    model.addFirst(job);
+                }
+            }
         }
     }
 }
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
@@ -16,17 +16,23 @@ import com.amazonaws.ml.mms.archive.ModelArchive;
 import com.amazonaws.ml.mms.archive.Signature;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class Model {
-
+    public static final String DEFAULT_DATA_QUEUE = "DATA_QUEUE";
     private ModelArchive modelArchive;
     private int minWorkers;
     private int maxWorkers;
     private int batchSize;
     private int maxBatchDelay;
-    private LinkedBlockingDeque<Job> jobs;
+    // Total number of subsequent inference request failures
+    private AtomicInteger failedInfReqs;
+    // Per workerthread job queue. This separates out the control queue from data queue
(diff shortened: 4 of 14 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/ModelLoadModelRequest.java
```diff
@@ -27,15 +27,15 @@
     private String gpu;
 
     public ModelLoadModelRequest() {
-        super("load", null);
+        super(CommandStrings.LOAD.getCommand(), null);
```
frankfliu: It's better use CommandStrings type in Request class as well.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/ModelLoadModelRequest.java
```diff
@@ -27,15 +27,15 @@
...
     }
 
     public ModelLoadModelRequest(Model model, String gpu) {
-        super(CommandStrings.LOAD, model.getModelName());
+        super(CommandStrings.LOAD.getCommand(), model.getModelName());
         this.handler = model.getModelArchive().getManifest().getModel().getHandler();
         this.batchSize = model.getBatchSize();
         this.gpu = gpu;
-        this.setIsSynchronous();
+        this.setSynchronous(true);
```
frankfliu: synchronous is already true in base class, you don't need set it.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/CommandStrings.java
```diff
@@ -1,10 +1,18 @@
 package com.amazonaws.ml.mms.util.messages;
 
-public final class CommandStrings {
-    public static final String PREDICT = "predict";
-    public static final String LOAD = "load";
-    public static final String UNLOAD = "unload";
-    public static final String STATS = "stats";
+public enum CommandStrings {
```
frankfliu: Better name this class as something like RequstCommand

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -118,12 +118,12 @@ public void setMaxBatchDelay(int maxBatchDelay) {
...
 
-    public boolean addJob(Long threadId, Job job) {
+    public boolean addJob(String threadId, Job job) {
         jobsDb.putIfAbsent(threadId, new LinkedBlockingDeque<>());
         return jobsDb.get(threadId).offer(job);
     }
 
-    public void removeJobQueue(Long threadId) {
+    public void removeJobQueue(String threadId) {
         if (!threadId.equals(DEFAULT_DATA_QUEUE)) {
```
frankfliu: Better to use if (!DEFAULT.equals(threadId))

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -82,6 +83,7 @@ public void modelChanged(Model model) throws WorkerInitializationException {
     private void addThreads(List<WorkerThread> threads, Model model, int count)
             throws WorkerInitializationException {
         int maxGpu = configManager.getNumberOfGpu();
+        UUID threadUuid = UUID.randomUUID();
```
frankfliu: This should be in for loop, otherwise all the worker has the same name,
Actually, each worker thread already has a unique name, you can just call thread.getName() to retrieve it.
And I believe your old way using threadId should work. You was using a wrong thread Id in addJob() call.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -96,21 +99,23 @@ public void run() {
         BaseModelRequest req = null;
         try {
             while (running.get()) {
-                req = aggregator.getRequest(currentThread.getId());
+                req = aggregator.getRequest(this.threadName);
```
frankfliu: use can use:  req = aggregator.getRequest(getName());

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/BaseModelRequest.java
```diff
@@ -16,20 +16,20 @@
 
     private String command;
     private String modelName;
-    private boolean isSynchronous;
+    private boolean synchronous;
```
frankfliu: In which case we need async?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
```diff
@@ -141,18 +126,29 @@ public void sendError(BaseModelRequest message, String error) {
...
+                Job job = jobs.remove(jobsId);
+                job.sendError(error);
             }
-            job.sendError(error);
         }
 
         if (!jobs.isEmpty()) {
-            throw new IllegalStateException("Not all jobs get response.");
+            jobs.clear();
+            logger.error("Huh!!! ");
```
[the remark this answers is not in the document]
frankfliu: Should we throw exception here if this thread need to be killed?

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/BaseModelRequest.java
```diff
@@ -16,10 +16,20 @@
 
     private String command;
     private String modelName;
+    private boolean synchronous;
```
frankfliu: In which case we use async?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
```diff
@@ -90,6 +74,7 @@ public BaseModelRequest getRequest() throws InterruptedException {
         ModelInferenceRequest req = new ModelInferenceRequest(model.getModelName());
         for (Job j : jobs.values()) {
             req.addRequestBatches(j.getPayload());
+            req.setSynchronous(true);
```
frankfliu: Default is true, redundant set

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/ModelLoadModelRequest.java
vdantu: Done

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
vdantu: True. I had added it to avoid any async messages triggering a failure. But we could adopt one of the two approaches.
1. Add this back when we have actual sync messages in the future to the backend.
2. Make backend worker always return a message (making it always a sync)

I will remove this for now, as its not needed.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
vdantu: Replaced this with getName().

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
```diff
@@ -141,18 +126,29 @@ public void sendError(BaseModelRequest message, String error) {
...
+            // Send the error message to all the jobs
+            for (Map.Entry<String, Job> j : jobs.entrySet()) {
+                String jobsId = j.getValue().getJobId();
+                Job job = jobs.remove(jobsId);
+                job.sendError(error);
             }
-            job.sendError(error);
         }
 
         if (!jobs.isEmpty()) {
```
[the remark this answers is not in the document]
frankfliu: In the if (message == null) block, you have looped through all jobs and already removed them. It must be empty.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -109,19 +118,59 @@ public void setMaxBatchDelay(int maxBatchDelay) {
...
+            retJob = jobsDb.get(tid).poll(timeout, TimeUnit.MILLISECONDS);
+        } else {
+            retJob = jobsDb.get(DEFAULT_DATA_QUEUE).poll(timeout, TimeUnit.MILLISECONDS);
+        }
+        return retJob;
     }
 
-    public Job nextJob() throws InterruptedException {
-        return jobs.take();
+    public int incrFailedInfReqs() {
```
frankfliu: we are not using the return value, should it be void?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -152,7 +168,7 @@ public void initChannel(Channel ch) {
                                         latch.countDown();
                                         parentThreads.remove(WorkerThread.this); // NOPMD
                                         shutdown();
-                                        logger.info("Worker disconnected.");
+                                        logger.info("{} Worker disconnected.", getName());
```
frankfliu: log4j default print out thread name.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -162,16 +178,28 @@ public void initChannel(Channel ch) {
...
                                         Job job =
-                                                new Job(null, model.getModelName(), "load", input);
-                                        model.addFirst(job);
+                                                new Job(
+                                                        null,
+                                                        model.getModelName(),
+                                                        CommandStrings.LOAD,
+                                                        input);
+                                        model.addJob(getName(), job);
+                                        latch.countDown();
```
frankfliu: Why countDown twice?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -201,8 +229,27 @@ public long getStartTime() {
...
+                if (parentThreads.size() < model.getMinWorkers()) {
+                    // minWorkers is the numWorkers configured for this model. If parent thread group
+                    // has less threads than expected minWorkers, we should restart the workers.
+                    manager.updateModel(
+                            model.getModelName(), model.getMinWorkers(), model.getMinWorkers());
+                }
+                aggregator.sendError(null, "Internal Failure");
+            } catch (WorkerInitializationException wie) {
+                logger.error(
+                        "Error restarting the backend worker for model name "
```
frankfliu: User var-arg error() method to avoid unnecessary string concat

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -201,8 +229,27 @@ public long getStartTime() {
...
         running.set(false);
         backendChannel.close();
+        ModelManager manager = ModelManager.getInstance();
         if (currentThread != null) {
             currentThread.interrupt();
+            try {
+                if (parentThreads.size() < model.getMinWorkers()) {
+                    // minWorkers is the numWorkers configured for this model. If parent thread group
+                    // has less threads than expected minWorkers, we should restart the workers.
+                    manager.updateModel(
```
frankfliu: Better add backoff on retry

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -201,8 +229,27 @@ public long getStartTime() {
...
+                            model.getModelName(), model.getMinWorkers(), model.getMinWorkers());
+                }
+                aggregator.sendError(null, "Internal Failure");
+            } catch (WorkerInitializationException wie) {
+                logger.error(
+                        "Error restarting the backend worker for model name "
+                                + model.getModelName());
+            }
+
+            if (manager.getModels().containsKey(model.getModelName())) {
```
frankfliu: getModelName() and getModels() called multiple times, better define variables for them

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -109,19 +118,59 @@ public void setMaxBatchDelay(int maxBatchDelay) {
...
+        jobsDb.get(threadId).addFirst(job);
     }
 
     public void addFirst(Job j) {
-        jobs.addFirst(j);
+        addFirst(DEFAULT_DATA_QUEUE, j);
+    }
+
+    public Job nextJob(String threadId) throws InterruptedException {
+        String tid = threadId != null ? threadId : DEFAULT_DATA_QUEUE;
```
frankfliu: It seems threadId is never null. There is no call path passing null threadId.
This can be simplified as:

    public Job nextJob(String threadId) throws InterruptedException {
        return nextJob(threadId, Long.MAX_VALUE);
    }

    public Job nextJob(String threadId, long timeout) throws InterruptedException {
        LinkedBlockingDeque<Job> jobs = jobsDb.get(threadId);
        if (jobs != null && !jobs.isEmpty()) {
            return jobs.poll();
        }
        return jobsDb.get(DEFAULT_DATA_QUEUE).poll(timeout, TimeUnit.MILLISECONDS);
    }

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
vdantu: I had removed this. I think its updated in the other PR. I will remove it from this PR as well.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: Sure. This looks more compact.

PR #456, opened 2018-07-19
Add code coverage report for gradle build.

frankfliu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ci/buildspec.yml
+++ b/ci/buildspec.yml
@@ -23,6 +23,7 @@ phases:
       - pip install -U -e .
   build:
     commands:
+      - frontend/gradlew -p frontend build
       - python -m pytest mms/tests/unit_tests
       # integration test is broken: https://github.com/awslabs/mxnet-model-server/issues/437
       #- python -m pytest mms/tests/integration_tests
--- a/frontend/build.gradle
+++ b/frontend/build.gradle
@@ -38,8 +38,19 @@ configure(javaProjects()) {
             events "passed", "skipped", "failed", "standardOut", "standardError"
         }
     }
+    test.finalizedBy(project.tasks.jacocoTestReport)
 
     compileJava {
         options.compilerArgs << "-Xlint:all,-options,-static" << "-Werror"
+    }
+
+    jacocoTestCoverageVerification {
+        violationRules {
+            rule {
+                limit {
+                    minimum = 0.65
+                }
+            }
+        }
     }
 }
--- a/frontend/build.gradle
+++ b/frontend/build.gradle
@@ -48,7 +48,7 @@ configure(javaProjects()) {
         violationRules {
             rule {
                 limit {
-                    minimum = 0.65
+                    minimum = 0.75
                 }
             }
         }
--- a/frontend/gradle/wrapper/gradle-wrapper.properties
+++ b/frontend/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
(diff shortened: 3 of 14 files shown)


frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/Manifest.java
```diff
@@ -190,20 +190,19 @@ public void setModelVersion(String modelVersion) {
...
 
-        public String getParametersFile() {
-            return parametersFile;
+        public Map<String, Object> getExtensions() {
+            return extensions;
         }
 
-        public void setParametersFile(String parametersFile) {
-            this.parametersFile = parametersFile;
+        public void setExtensions(Map<String, Object> extensions) {
```
piyushghai: Should we have a null check here as well ? and if a null extensions is passed, we set extensions to a new instance of LinkedHashMap ?

frankfliu: The extensions is nullable. If extensions is null, we don't serialize this field into json. extensions will be created if addExtensions() is called at later time.

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/Manifest.java
```diff
@@ -190,20 +190,19 @@ public void setModelVersion(String modelVersion) {
...
-            return symbolFile;
-        }
-
-        public void setSymbolFile(String symbolFile) {
-            this.symbolFile = symbolFile;
+        public void addExtension(String key, Object value) {
+            if (extensions == null) {
+                extensions = new LinkedHashMap<>();
+            }
+            extensions.put(key, value);
```
vdantu: Doesn't this rewrite a key's value, if the key already exists? Should we throw warning in case where users are unknowingly creating duplicate key-value extensions?

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
```diff
@@ -78,7 +78,7 @@ public static ModelArchive downloadModel(String modelStore, String url)
             }
         }
 
-        if (url.startsWith(".")) {
+        if (url.contains("..")) {
```
vdantu: What is this used for? Is this a "relative paths" check?

frontend/modelarchive/src/test/java/com/amazonaws/ml/mms/archive/Exporter.java
```diff
@@ -121,17 +121,17 @@ public static void main(String[] args) {
...
             File symbolFile = findUniqueFile(files, "-symbol.json");
             if (symbolFile != null) {
-                model.setSymbolFile(symbolFile.getName());
+                model.addExtension("symbolFile", symbolFile.getName());
             }
 
             File paramsFile = findUniqueFile(files, ".params");
             if (paramsFile != null) {
-                model.setParametersFile(paramsFile.getName());
+                model.addExtension("parametersFile", paramsFile.getName());
```
vdantu: IS there a way for customers to feed in the symbol/params file? This seems like we are assuming we will have only one params/symbol file or else we throw an exception.

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
frankfliu: If the url contains .., hacker can navigate to any file path.

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/Manifest.java
frankfliu: This is a setter like method, it doesn't make sense to check duplicate here.
How can you throw a warning here?
How do you know if user want to override the value of not?

frontend/modelarchive/src/test/java/com/amazonaws/ml/mms/archive/Exporter.java
frankfliu: 1. This is an unit test code.
2. This logic is the same as current python export logic. I think we should allow user to specify these values, but I preferred do it with a UI.

frankfliu:
This PR contains 4 commits, they are mixed with different changes.

vrakesh:
Test coverage part looks good , we can merge this once conflict is resolved.

PR #460, opened 2018-07-23
Ping changes

vdantu:
*Issue #, if available:*

*Description of changes:*
"Ping" to return more information about the system. 
1. Added async update of number of workers running and number of workers configured
1. Ping command will run as a thread pool and return the response to avoid http-handler threads from being blocked.

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
@@ -120,7 +120,7 @@ public class HttpRequestHandler extends SimpleChannelInboundHandler<FullHttpRequ
     }
 
     private void handlePing(ChannelHandlerContext ctx) {
-        NettyUtils.sendJsonResponse(ctx, new StatusResponse("healthy"));
+        ModelManager.getInstance().workerStatus(ctx);
     }
 
     private void handleModelsApi(
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ErrorResponse.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ErrorResponse.java
@@ -18,8 +18,7 @@ public class ErrorResponse {
     private String type;
     private String message;
 
-    public ErrorResponse() {
-    }
+    public ErrorResponse() {}
 
     public ErrorResponse(int code, String message) {
         this.code = code;
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
@@ -16,7 +16,10 @@ import com.amazonaws.ml.mms.archive.InvalidModelException;
 import com.amazonaws.ml.mms.archive.Manifest;
 import com.amazonaws.ml.mms.archive.ModelArchive;
 import com.amazonaws.ml.mms.common.ErrorCodes;
+import com.amazonaws.ml.mms.http.StatusResponse;
 import com.amazonaws.ml.mms.util.ConfigManager;
+import com.amazonaws.ml.mms.util.NettyUtils;
+import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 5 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
```diff
@@ -127,7 +129,22 @@ private void handleListModels(ChannelHandlerContext ctx, FullHttpRequest req) {
     }
 
     private void handlePing(ChannelHandlerContext ctx) {
-        NettyUtils.sendJsonResponse(ctx, new StatusResponse("healthy"));
+        Map<String, String> status = new HashMap<>();
+        Map<String, Model> models = ModelManager.getInstance().getModels();
```
frankfliu: This list may become large. It's better we only return overall status.
We have list models and describe model api, user can use that the retrieve each models status.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
```diff
@@ -127,7 +129,22 @@ private void handleListModels(ChannelHandlerContext ctx, FullHttpRequest req) {
...
+        HttpResponseStatus ret = HttpResponseStatus.OK;
+        for (Map.Entry<String, Model> m : models.entrySet()) {
+            int numScaled = m.getValue().getMinWorkers();
+            int numWorking = m.getValue().getCurrentWorkers().get();
+            String s = numScaled + "/" + numWorking;
+            if (numScaled > 0) {
```
frankfliu: This might be problematic if the model thread is restarting.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -44,6 +45,15 @@ public Model(ModelArchive modelArchive, int queueSize) {
...
         jobsDb.putIfAbsent(DEFAULT_DATA_QUEUE, new LinkedBlockingDeque<>(queueSize));
         failedInfReqs = new AtomicInteger(0);
+        currentWorkers = new AtomicInteger(0);
+    }
+
+    public AtomicInteger getCurrentWorkers() {
+        return currentWorkers;
+    }
+
+    public void setCurrentWorkers(AtomicInteger currentWorkers) {
```
frankfliu: It doesn't seem right have a getter/setter for currentWorkers.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
```diff
@@ -127,7 +129,22 @@ private void handleListModels(ChannelHandlerContext ctx, FullHttpRequest req) {
...
+                ret = HttpResponseStatus.NOT_FOUND;
+                if (numWorking > 0) {
+                    ret = HttpResponseStatus.OK;
+                }
+            }
+            status.put(m.getKey(), s);
+        }
+        NettyUtils.sendJsonResponse(ctx, new StatusResponse(JsonUtils.GSON.toJson(status)), ret);
```
frankfliu: The response is always 200, so the checking of models status doesn't add any value to this API.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: True. I will remove this setter. I think get should be enough.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
vdantu: It does seem a lot. Question is do we foresee handling 1000's of models at once. If not, this should be OK I think. In any case, I can change this and wait on feedback from customers on if this is useful.
For now I will return one of the two error codes 200 (OK) and 206 (Partial content) to distinguish between all the workers are running and only some of the workers are running, in case of SUCCESS. I will make that change.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
vdantu: That should be OK. The idea here is that we display how many they assigned and how many they have running and ready to serve. This is not used for scaling, customers will be asked to use the "list" command to scale and that will have all the state info.

PR #461, opened 2018-07-23
Unit testing for client sdk generator

zachgk:
Fix spelling error lanugage->language in clientSDKGenerator
Use with notation for file opening

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
@@ -120,7 +120,7 @@ public class HttpRequestHandler extends SimpleChannelInboundHandler<FullHttpRequ
     }
 
     private void handlePing(ChannelHandlerContext ctx) {
-        NettyUtils.sendJsonResponse(ctx, new StatusResponse("healthy"));
+        ModelManager.getInstance().workerStatus(ctx);
     }
 
     private void handleModelsApi(
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
@@ -16,7 +16,10 @@ import com.amazonaws.ml.mms.archive.InvalidModelException;
 import com.amazonaws.ml.mms.archive.Manifest;
 import com.amazonaws.ml.mms.archive.ModelArchive;
 import com.amazonaws.ml.mms.common.ErrorCodes;
+import com.amazonaws.ml.mms.http.StatusResponse;
 import com.amazonaws.ml.mms.util.ConfigManager;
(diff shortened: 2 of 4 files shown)


piyushghai:
I think this commit : https://github.com/awslabs/mxnet-model-server/pull/462/commits/6e6f6d595b1b3a728ce268f63380d4fbd2f5e958  will fix the CI for your PR

mms/tests/unit_tests/test_client_sdk_generator.py
```diff
@@ -0,0 +1,66 @@
...
+    return patches
+
+def test_handles_exception(patches):
+    test_exception = Exception("test")
+    patches.path_exists.side_effect = test_exception
+    with pytest.raises(Exception) as excinfo:
+        ClientSDKGenerator.generate(sample_openapi_endpoints, sample_sdk_language)
+    patches.json_dump.assert_not_called()
+    patches.subprocess_call.assert_not_called()
+    assert excinfo.value != test_exception
```
piyushghai: Should we actually be comparing against a dummy object ? Should we rather not compare the error message we are getting back from the exception ?

zachgk: I don't want to copy the contents of the error message text in the test.  What really should happen (and we do elsewhere), is that we use the MSError class.  Then we can check whether the exception returned is an MSError and potentially that it uses the correct error code.  Here, I just ensure that the exception was handled in some fashion because it raises a new exception

zachgk: I don't want to copy the contents of the error message text in the test. What really should happen (and we do elsewhere), is that we use the MSError class. Then we can check whether the exception returned is an MSError and potentially that it uses the correct error code. Here, I just ensure that the exception was handled in some fashion because it raises a new exception

mms/client_sdk_generator.py
```diff
@@ -25,34 +25,33 @@ class ClientSDKGenerator(object):
     Class for client SDK Generator
     """
     @staticmethod
-    def generate(openapi_endpoints, sdk_lanugage):
+    def generate(openapi_endpoints, sdk_language):
```
frankfliu: We should revisit this:
1. We have upgraded to openapi 3.0, we should get rid of swager 2.0
2. the json will be generated by java code. we should use openapi library to create json file.
3. It's better have a single java entry to create a json and call swager 3.0 to generate the client package
4. We should revisit the use cases:
a. only generate client for single model entry point
b. generate only model management APIs
c. should we generate deprecated APIs?
d. Can we generate multiple models APIs?

ci/Dockerfile.python3.6
```diff
@@ -190,6 +190,7 @@ RUN set -ex \
     && pip install retrying \
     && pip install mock \
     && pip install pytest -U \
+    && pip install pytest-mock -U \
```
frankfliu: Please move this into buildspec.yml.

PR #458, opened 2018-07-20
Unit tests for model_loader.py

piyushghai:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #463, opened 2018-07-24
Unit tests for model_loader.py

piyushghai:
This PR is a replacement for #458 

### DIFF
--- a/mms/tests/unit_tests/test_model_loader.py
+++ b/mms/tests/unit_tests/test_model_loader.py
@@ -1,0 +1,151 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import pytest
+import os
+import shutil
+import json
+from mms.model_loader import MANIFEST_FILENAME
+from mms.model_loader import ModelLoader
+
+model_dir_path = 'my-model'
+handler_file = 'handler'
+
+manifest_invalid_data_missing_extensions = {"model": {}, "engine": {"engineName": "MxNet"}}
+manifest_invalid_data_missing_paramsFile = {"model": {"extensions" : {}}, "engine": {"engineName": "MxNet"}}
+manifest_invalid_data_missing_symbolFile = {"model": {"extensions:": {"parametersFile": "my-model/params1"}},
+                                            "engine": {"engineName": "MxNet"}}
+manifest_valid_data = {"model": {"extensions": {"parametersFile": "my-model/params1", "symbolFile": 'my-model/symbol.json'}},
+                       "engine": {"engineName": "MxNet"}}
+
+
+def empty_file(filepath):
+    """
+    Utility function for creating an empty file, which will be used throughout these unit tests
+    :param filepath:
+    :return:
+    """
+
+    open(filepath, 'a').close()
+
+
+@pytest.mark.skip(reason="Disabling it currently until the PR #467 gets merged")
+class TestModelLoad():
+
+    @pytest.fixture(scope='session')
+    def create_empty_manifest_file(self):
+        """
+        By setting the scope of create_empty_manifest_file as session we are ensuring that this piece of code runs once for all of the tests combined in this file
+        """
+        # Setup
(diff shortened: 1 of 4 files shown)


mms/tests/unit_tests/test_model_loader.py
```diff
@@ -0,0 +1,155 @@
...
+
+@pytest.mark.usefixtures('create_empty_manifest_file')
+def test_manifest_file_open():
+    with pytest.raises(Exception):
+        ModelLoader.load(model_dir_path, handler_file)
+
+
+@pytest.mark.usefixtures('create_empty_manifest_file')
+def test_parameter_file_defined_in_manifest():
+    with open(model_dir_path + '/' + MANIFEST_FILENAME, 'w') as f:
```
vrakesh: string join or os path join

mms/utils/validators/validate_messages.py
```diff
@@ -35,10 +35,13 @@ def validate_load_message(msg):
         :return:
         """
         if u'modelPath' not in msg:
```
vrakesh: unicode still required??? builtins should handle it

mms/tests/unit_tests/test_validate_messages.py
```diff
@@ -15,25 +15,35 @@
...
 
 def test_validate_load_message_missing_model_path():
-    invalid_object = {'command': 'some-command', 'modelName': 'some-model-name'}
+    invalid_object = {'command': 'some-command', 'modelName': 'some-model-name', 'handler': 'some handler string'}
     with pytest.raises(MMSError) as error:
         ModelWorkerMessageValidators.validate_load_message(invalid_object)
 
     assert error.value.get_code() == ModelServerErrorCodes.INVALID_LOAD_MESSAGE, 'error codes don\'t match'
-    assert error.value.get_message() == 'Load command missing \"model-path\" key'
+    assert error.value.get_message() == 'Load command missing \"modelPath\" key'
```
vrakesh: Double qoute string to avoid using escape sequences
Eg:
  "Hello I'm here"
Double quote strings, is a better way always

mms/tests/unit_tests/test_model_loader.py
```diff
@@ -0,0 +1,155 @@
...
+
+def empty_file(filepath):
+
+    """
+    Utility function for creating an empty file, which will be used throughout these unit tests
+    :param filepath:
+    :return:
+    """
+
+    open(filepath, 'a').close()
```
zachgk: If we are using mocker, we probably want to use https://docs.python.org/dev/library/unittest.mock.html#mock-open to avoid messing with the filesystem.

mms/tests/unit_tests/test_model_loader.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@pytest.mark.usefixtures('create_empty_manifest_file')
+def test_manifest_file_open():
+    with pytest.raises(Exception):
+        ModelLoader.load(model_dir_path, handler_file)
+
+
+@pytest.mark.usefixtures('create_empty_manifest_file')
+def test_parameter_file_defined_in_manifest():
```
zachgk: You can use the usefixtures by parameter shorthand

mms/tests/unit_tests/test_model_loader.py
```diff
@@ -0,0 +1,155 @@
...
+def create_empty_manifest_file():
+    """
+    By setting the scope of create_empty_manifest_file as session we are ensuring that this piece of code runs once for all of the tests combined in this file
+    """
+    # Setup
+    path = '{}/'.format(model_dir_path)
+    if os.path.exists(path):
+        shutil.rmtree(path)
+    os.mkdir(path)
+    path = '{}/'.format(path) + MANIFEST_FILENAME
```
zachgk: You could use '{}/{}' or os.path.join.  There are also other places in the file where os.path.join should be used.

mms/tests/unit_tests/test_model_loader.py
piyushghai: Good point. Will do it.

mms/tests/unit_tests/test_model_loader.py
piyushghai: Will do it.

mms/tests/unit_tests/test_model_loader.py
piyushghai: Discussed offline. Skipping the mock for os funcs for now.

mms/utils/validators/validate_messages.py
piyushghai: This is used everywhere. So I'm leaving it there for now. 
Changing the unicode extension can be a part of a separate refactoring of this file.

mms/tests/unit_tests/test_validate_messages.py
piyushghai: Done in eaa62a366e264cec1e8290a247a17f279b31fa63

PR #469, opened 2018-07-27
Add Unit Tests to model_service_worker

zachgk:
Issue 759

*Description of changes:*

### DIFF
--- a/mms/model_service_worker.py
+++ b/mms/model_service_worker.py
@@ -22,6 +22,7 @@ import json
 from builtins import bytes
 from builtins import str
 
+from mms.model_loader import ModelLoader
 from mms.service_manager import ServiceManager
 from mms.log import log_msg, log_error
 from mms.utils.validators.validate_messages import ModelWorkerMessageValidators
@@ -31,7 +32,7 @@ from mms.utils.model_server_error_codes import ModelServerErrorCodes as err
 from mms.metrics.metric_encoder import MetricEncoder
 
 MAX_FAILURE_THRESHOLD = 5
-
+debug = False
 
 class MXNetModelServiceWorker(object):
     """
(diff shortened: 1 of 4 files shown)


ci/Dockerfile.python2.7
```diff
@@ -179,6 +179,7 @@ RUN set -ex \
     && pip install retrying \
     && pip install mock \
     && pip install pytest -U \
+    && pip install pytest-mock -U \
```
piyushghai: These shouldn't be there. The pytest-mock install should go in buildspec.yml

mms/tests/unit_tests/test_model_service_worker.py
```diff
@@ -0,0 +1,606 @@
...
+
+            sock.close.assert_called()
+            patches.log.assert_called_with(log_call_param)
+
+        def test_stop_server(self, patches, worker, sock):
+            worker.stop_server(sock)
+            worker.send_response.assert_called()
+            sock.close.assert_called()
+
+
```
piyushghai: Too many extra lines here.

mms/tests/unit_tests/test_model_service_worker.py
```diff
@@ -0,0 +1,606 @@
...
+
+
+def test_emit_metrics(mocker):
+    metrics = {'test_emit_metrics': True}
+    printer = mocker.patch('mms.model_service_worker.print')
+    dumps = mocker.patch('json.dumps')
+    flush = mocker.patch('sys.stdout')
```
piyushghai: What are we doing with the flush mocker ?

mms/model_service_worker.py
```diff
@@ -91,10 +92,10 @@ def create_predict_response(self, ret, req_id_map, invalid_reqs):
                 result.update({"requestId": req_id_map[idx]})
                 result.update({"code": 200})
 
-                if isinstance(val, bytes):
-                    value = ModelWorkerCodecHelper.encode_msg(encoding, val)
-                elif isinstance(val, str):
+                if isinstance(val, str):
```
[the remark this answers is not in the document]
zachgk: In python2, str and bytes are considered the same.  We switched them while investigating this and attempting to achieve full coverage.  Ultimately, python2 will register the second option uncovered during testing regardless of the order, but it does at least return the correct result.

PR #478, opened 2018-08-02
Pip package for MMS 1.0 alpha

vdantu:
*Issue #, if available:*

*Description of changes:*
Pip package is available at test pypi for MMS 1.0

pip install -U --index-url https://test.pypi.org/simple/  --pre --extra-index-url https://pypi.org/simple mxnet-model-server --no-cache-dir

run 
mxnet-model-server -h for how to run the mms


mms/model_service_worker.py
```diff
@@ -37,10 +37,14 @@ class MXNetModelServiceWorker(object):
     """
     Backend worker to handle Model Server's python service code
     """
-    def __init__(self, s_name=None):
-        if s_name is None:
-            raise ValueError("Incomplete data provided: Model worker expects \"socket name\"")
-        self.sock_name = s_name
+    def __init__(self, s_type=None, s_name=None):
```
vdantu: Ignore this. Its from another PR.

mms/model_service_worker.py
```diff
@@ -17,11 +17,11 @@
```
vdantu: Ignore this file for this PR

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -17,6 +17,8 @@
 import java.io.File;
```
vdantu: Ignore this file for this PR. Its being tracked by another PR.

setup.py
```diff
@@ -30,44 +16,138 @@ def PyPiDescription():
...
+dest_file_name = os.path.abspath('mms/frontend/model-server.jar')
+
+
+def pypi_description():
+    """Imports the long description for the project page"""
+    with open('PyPiDescription.rst') as df:
+        return df.read()
+
+
+def detect_model_server_version():
```
frankfliu: Version should be passed as argument during the build.
If version is not provided, then use default version

setup.py
```diff
@@ -30,44 +16,138 @@ def PyPiDescription():
...
+    with open(os.path.abspath(os.path.join("mms", "version.py")), 'r') as vf:
+        exec(vf.read(), None, globals())
+    return __version__
+
+
+def select_mxnet(requirements=[]):
```
frankfliu: This should be passed as parameter as well, should not depend on current host.

Also, "requirements" shadows variable in outer scope. You don't really need pass it in, you can append the return value at calling function.

setup.py
```diff
@@ -30,44 +16,138 @@ def PyPiDescription():
...
+    if platform.system().lower() == 'linux':
+        try:
+            # Check if CUDA is installed
+            cuda = ctypes.cdll.LoadLibrary('libcudart.so')
+            device_count = ctypes.c_int()
+            # get the number of supported GpUs
+            cuda.cudaGetDeviceCount(ctypes.byref(device_count))
+            if device_count.value > 0:
+                gpu_platform = True
+        except Exception as e:
```
frankfliu: Local variable e is not used

setup.py
```diff
@@ -30,44 +16,138 @@ def PyPiDescription():
...
+        name='mxnet-model-server',
+        version=version.strip()+'a2',
+        description='Model Server for Apache MXNet is a tool for serving neural net models for inference',
+        long_description=pypi_description(),
+        url='https://github.com/awslabs/mxnet-model-server',
+        keywords='MXNet Model Server Serving Deep Learning Inference AI',
+        packages=pkgs,
+        cmdclass={
+            'build_frontend': BuildFrontEnd,
+            'build_py':BuildPy,
```
frankfliu: Missing space

setup.py
```diff
@@ -30,44 +16,138 @@ def PyPiDescription():
...
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        try:
+            front_end_bin_dir = os.path.abspath('.') + '/mms/frontend'
```
frankfliu: os.path.abspath('.') might throw exception, and will cause crash in exception handling code, due to front_end_bin_dir has not been initialized.

setup.py
vdantu: Yes, we will need to look into this a little more in detail. I would like to take it up as a part of putting it together in the pipeline.

setup.py
vdantu: True. This can be removed.

setup.py
vdantu: True. This can be removed as well.

setup.py
vdantu: True added it. I should probably look into attaching a pylint for this file as well.

setup.py
vdantu: The assumption with setup.py is we run python build in the same directory where setup.py exists. If there is an exception, it is raised to the above level and the build for setup.py will fail.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -16,15 +16,16 @@
 import com.amazonaws.ml.mms.util.ConfigManager;
 import com.amazonaws.ml.mms.util.JsonUtils;
 import com.amazonaws.ml.mms.util.NettyUtils;
+<<<<<<< HEAD
```
piyushghai: Can you remoe the merge conflicts here ?

mms/model_service_worker.py
```diff
@@ -21,7 +21,10 @@
 from builtins import bytes
 from builtins import str
 
+<<<<<<< HEAD
 from mms.model_loader import ModelLoader
+=======
```
piyushghai: Merge conflicts.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
piyushghai: Oh they are removed in a separate commit.  Just saw.

mms/model_service_worker.py
piyushghai: Oh they are removed in a separate commit. Just saw.

PR #485, opened 2018-08-10
Dockerfile changes

vdantu:
*Issue #, if available:*
*Description of changes:*
1. Docker file changes for the MMS 1.0
2. Model Server start/stop script changes


docker/Dockerfile.cpu
```diff
@@ -1,12 +1,26 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && cd /tmp \
+    && curl -O https://bootstrap.pypa.io/get-pip.py \
+    && python get-pip.py
 
-COPY mms_app_cpu.conf wsgi.py setup_mms.py mxnet-model-server /mxnet_model_server/
+RUN sudo pip install mxnet-model-server
 
-ENV PATH="/mxnet_model_server:${PATH}" MXNET_MODEL_SERVER_GPU_IMAGE=0
+RUN useradd -m docker && echo "docker:docker" | chpasswd && adduser docker sudo
```
frankfliu: should we use mms (or: model-server) as user name instead of docker?

frankfliu: Why this user need to be a sudoer?

docker/Dockerfile.cpu
```diff
@@ -1,12 +1,26 @@
...
-RUN pip install mxnet-model-server \
-    && pip uninstall --yes mxnet \
-    && pip install mxnet-mkl \
-    && mkdir /mxnet_model_server
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    build-essential \
+    libatlas-base-dev \
+    libopencv-dev \
+    graphviz \
```
frankfliu: Why we need all these but without mxnet? In future mms will not install mxnet.

mms/model_server.py
```diff
@@ -20,19 +20,13 @@ def start():
...
-            with open(pid_file, 'r') as f:
-                try:
-                    os.kill(int(f.readline()), signal.SIGKILL)
-                except OSError:
-                    print("Model server already stopped")
-            os.remove(pid_file)
-        else:
-            print("Model server is not currently running")
+    else:  # args.stop is True:
+        # TODO: Can we write this in a better way?
```
frankfliu: Why we move away from use pid file?

docker/Dockerfile.cpu
vdantu: Oh, regarding why MXNet is not a part of this docker, its because its a part of "mxnet-model-server"s pip install

PR #489, opened 2018-08-15
Fix for Legacy Manifest handling issue

piyushghai:
*Issue #, if available:*

*Description of changes:*
Added handling for reading Manifest.legacy file, if present and passing it if not None to custom service code's class def method. 

### DIFF
--- a/mms/model_loader.py
+++ b/mms/model_loader.py
@@ -18,6 +18,7 @@ from mms.log import get_logger
 logger = get_logger()
 
 MANIFEST_FILENAME = 'MANIFEST.json'
+MANIFEST_LEGACY_FILE = 'MANIFEST.legacy'
 
 
 class ModelLoader(object):
@@ -43,10 +44,21 @@ class ModelLoader(object):
         """
         manifest = None
         manifest_file = os.path.join(model_dir, MANIFEST_FILENAME)
+
+        # Legacy manifest handling
+        manifest_legacy = None
+        manifest_legacy_file = os.path.join(model_dir, MANIFEST_LEGACY_FILE)
+
+        if os.path.isfile(manifest_legacy_file):
+            try:
+                manifest_legacy = json.load(open(manifest_legacy_file))
+            except ValueError as e:
+                raise Exception('Failed to open legacy manifest file. Stacktrace : ' + repr(e))
+
         if os.path.isfile(manifest_file):
             try:
                 manifest = json.load(open(manifest_file))
-            except Exception as e:
+            except ValueError as e:
                 raise Exception('Failed to open manifest file. Stacktrace: ' + str(e))
             model = manifest['model']
             engine = manifest['engine']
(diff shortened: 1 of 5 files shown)


mms/model_loader.py
```diff
@@ -43,6 +44,17 @@ def load(model_dir, handler):
...
         manifest_file = os.path.join(model_dir, MANIFEST_FILENAME)
+
+        # Legacy manifest handling
+        manifest_legacy = None
+        manifest_legacy_file = os.path.join(model_dir, MANIFEST_LEGACY_FILE)
+
+        if os.path.isfile(manifest_legacy_file):
+            try:
+                manifest_legacy = json.load(open(manifest_legacy_file))
+            except Exception as e:
```
vrakesh: generic exception make less generic

mms/service_manager.py
```diff
@@ -98,7 +98,7 @@ def get_loaded_modelservices(self, model_names=None):
             for model_name in model_names
         }
 
-    def load_model(self, model_name, model_dir, manifest, model_service_class_def, gpu=None, batch_size=None):
+    def load_model(self, model_name, model_dir, manifest, model_service_class_def, gpu=None, batch_size=None, manifest_legacy=None):
```
zachgk: I would put manifest_legacy in the middle of the parameters right after manifest.  It groups the manifest arguments together and allows easier utilization of the optional arguments. Also for register_and_load_modules.

mms/service_manager.py
```diff
@@ -119,6 +119,9 @@ def load_model(self, model_name, model_dir, manifest, model_service_class_def, g
         batch_size : int
             batch size
         """
+        # Check which manifest to pass down
+        manifest = manifest if manifest_legacy is None else manifest_legacy
```
zachgk: If manifest legacy is used the same way as manifest, you may want to move this line to ModelLoader and only pass the existing manifest file through all the various calls.  It would be much simpler and require less code.  But, does legacy not require special handling?

mms/model_loader.py
```diff
@@ -43,10 +44,21 @@ def load(model_dir, handler):
...
+
+        # Legacy manifest handling
+        manifest_legacy = None
+        manifest_legacy_file = os.path.join(model_dir, MANIFEST_LEGACY_FILE)
+
+        if os.path.isfile(manifest_legacy_file):
+            try:
+                manifest_legacy = json.load(open(manifest_legacy_file))
+            except ValueError as e:
+                logger.info(repr(e))
```
vdantu: This shouldn't be info it should be Exception. Can we assume service code will work without the legacy code?

mms/model_loader.py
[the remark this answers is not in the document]
piyushghai: Oddly enough, it passed here, even though we don't have a pylint disable line. 
https://github.com/awslabs/mxnet-model-server/blob/a4a8527bba17aed57079898a10eb8bcf72337832/mms/model_loader.py#L49

mms/model_loader.py
piyushghai: Oh right. It might not work without legacy files. I see your point. Will raise an exception here.

piyushghai: Done.

PR #490, opened 2018-08-16
Added Wheel build-artifact to CI

vrakesh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ci/buildspec.yml
+++ b/ci/buildspec.yml
@@ -7,20 +7,17 @@ phases:
     commands:
       - apt-get update
       - apt-get install -y curl
-      - apt-get install -y libgfortran3
-      - pip install future
       - pip install pip -U
-      - pip install flask
-      #- pip install -U mxnet
+      - pip install future
       - pip install Pillow
-      - pip install requests
-      - pip install flask-cors
-      - pip install fasteners
-      - pip install jsonschema
+      - pip install pytest
       - pip install pytest-mock -U
-      # - pip install onnx -U
-      # - pip install onnx-mxnet -U
-      - pip install -U -e .
+      - pip install -U -e .[mxnet]
+
+  pre_build:
+    commands:
+        - $NIGHTLYBRANCH
+            
   build:
     commands:
       - frontend/gradlew -p frontend build
@@ -28,8 +25,10 @@ phases:
       # integration test is broken: https://github.com/awslabs/mxnet-model-server/issues/437
       #- python -m pytest mms/tests/integration_tests
       - pylint -rn --rcfile=./mms/tests/pylintrc mms/.
+      - $NIGHTLYBUILD
+      - $NIGHTLYUPLOAD
 
 # local build will hang without artifacts section
 artifacts:
   files:
-    - README.md
+    - dist/*.whl
--- a/setup.py
+++ b/setup.py
@@ -32,9 +32,9 @@ import platform
 import os
 import ctypes
 import errno
-import time
+from datetime import date
 import sys
-from shutil import copyfile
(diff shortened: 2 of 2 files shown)


ci/buildspec.yml
```diff
@@ -28,8 +28,9 @@ phases:
       # integration test is broken: https://github.com/awslabs/mxnet-model-server/issues/437
       #- python -m pytest mms/tests/integration_tests
       - pylint -rn --rcfile=./mms/tests/pylintrc mms/.
+      - python setup.py bdist_wheel --universal
```
vdantu: This wouldn't work in CI. Same version can't be uploaded to PyPi. I would suggest testing these changes on "test PyPi" before pushing PR.

frankfliu: Can we remove unnecessary dependencies?
1.  apt-get install -y libgfortran3?
2. pip install flask
3. pip install flask-cors
4. pip install requests
5. pip install jsonschema
6. pip install -U -e . ?

frankfliu: And also in setup.py seems requires "importlib2", how our build work without that library?

setup.py
```diff
@@ -57,36 +57,27 @@ def detect_model_server_version():
     return __version__
 
 
-def add_requirements_to_2_x(pkg_name):
-    import sys
-    if sys.version_info[0] < 3:
-        requirements.append(pkg_name)
-
-
 def select_mxnet():
```
vdantu: Wondering if we should move "mxnet" pacakge to "extras_requires" .. This might be a better way to install mxnet at runtime. 
```
...
setup()
...
        extras_require={
            'mx_cpu': ['mxnet-mkl==1.2.0'],
            'mx_gpu': ['mxnet-cu90mkl>=1.2'],
        },
...
```
This way, when we install we could ask customers to install as
```
# MMS on CPU
$ pip install -U mxnet-model-server[mx_cpu]

# MMS on GPU
$ pip install -U mxnet-model-server[mx_gpu]
```

Can't think of a better way to install mxnet version based on CPU . vs GPU.

setup.py
```diff
@@ -32,16 +32,21 @@
...
 from setuptools import setup, find_packages, Command
 import setuptools.command.build_py
 
 pkgs = find_packages()
 source_server_file = os.path.abspath('frontend/server/build/libs/server-1.0.jar')
 dest_file_name = os.path.abspath('mms/frontend/model-server.jar')
+extras = {
+    'mxnet_cpu': ['mxnet-mkl==1.2.0'],
+    'mxnet_gpu': ['mxnet-cu90mkl==1.2.0'],
+    'mxnet': ['mxnet==1.2'],
```
vdantu: Whats the difference between this and mxnet_cpu?

setup.py
```diff
@@ -32,16 +32,21 @@
...
-from shutil import copyfile
+from shutil import copyfile, rmtree
 import subprocess
 from setuptools import setup, find_packages, Command
 import setuptools.command.build_py
 
 pkgs = find_packages()
 source_server_file = os.path.abspath('frontend/server/build/libs/server-1.0.jar')
 dest_file_name = os.path.abspath('mms/frontend/model-server.jar')
+extras = {
```
vdantu: Also, could you move this to the 'extras_require' ? Why is this a global?

setup.py
vrakesh: mxnet-mkl is intel mkl
vanilla mxnet uses openBLAS which works better on AMD cpus correct that is the difference correct?

setup.py
vrakesh: will move it in

PR #503, opened 2018-08-30
Timeout for backend worker accept

zachgk:
Add a 30 second timeout for the backend workers waiting for connection from frontend

### DIFF
--- a/mms/model_service_worker.py
+++ b/mms/model_service_worker.py
@@ -32,6 +32,7 @@ from mms.utils.model_server_error_codes import ModelServerErrorCodes as err
 from mms.metrics.metric_encoder import MetricEncoder
 
 MAX_FAILURE_THRESHOLD = 5
+SOCKET_ACCEPT_TIMEOUT = 30.0
 debug = False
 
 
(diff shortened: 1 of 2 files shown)


mms/model_service_worker.py
```diff
@@ -453,11 +454,16 @@ def run_server(self):
             try:
                 log_msg("Waiting for a connection")
 
+                self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
                 (cl_socket, _) = self.sock.accept()
+                self.sock.settimeout(socket.getdefaulttimeout())
```
frankfliu: Why this is needed?

zachgk: Otherwise, it sets a 30s timeout for all socket operations which I believe includes listening and writing

vdantu: self.sock and cl_socket are different sockets. You would get a new socket after you accept a connection. self.sock is only used for listening and accepting new connections. Do you mean cl_socket is inheriting this property from the self.sock?

zachgk: The documentation does not say that it will inherit the timeout, so cl_socket should be fine

PR #511, opened 2018-09-06
RequestId always returned

zachgk:
*Issue #, if available:*

*Description of changes:*


frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/RequestBatch.java
```diff
@@ -23,7 +23,11 @@
     private List<ModelInputs> modelInputs;
 
     public RequestBatch() {
```
frankfliu: Better remove default constructor

PR #513, opened 2018-09-06
Fix connection time bug for OSX

frankfliu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/mms/model_service_worker.py
+++ b/mms/model_service_worker.py
@@ -442,6 +442,9 @@ class MXNetModelServiceWorker(object):
         :return:
         """
         try:
+            self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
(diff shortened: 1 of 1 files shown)


piyushghai:
Thanks for fixing this :)

mms/model_service_worker.py
```diff
@@ -442,6 +442,9 @@ def run_server(self):
         :return:
         """
         try:
+            self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
+            self.sock.setblocking(True)  # workaround error(35, 'Resource temporarily unavailable') on OSX
```
vdantu: Is it by set to non-blocking after setting a timeout?

frankfliu: Some how it messed up socket opt in some version of python on OSX.

PR #518, opened 2018-09-12
Fix Backend worker socket not honoring timeout

vrakesh:
*Issue #, if available:*
The issue was caused by the fact that
calling socket.setblocking(True) is equivalent to saying socket.settimeout(None)
So we need to set the timeout after setting the blocking flag.
reference
https://docs.python.org/2/library/socket.html
*Description of changes:*

### DIFF
--- a/mms/model_service_worker.py
+++ b/mms/model_service_worker.py
@@ -178,7 +178,6 @@ class MXNetModelServiceWorker(object):
         """
         try:
             self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
-            self.sock.setblocking(True)  # workaround error(35, 'Resource temporarily unavailable') on OSX
 
             if self.sock_type == 'unix':
                 self.sock.bind(self.sock_name)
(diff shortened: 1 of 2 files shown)


piyushghai:
Can you indicate the steps taken to test it ?

vrakesh:
Sure, ran the script from command line, setting the timeout to 5.0 and later 3.0, with the statements ordered correctly, the timeout occurred without any issues. I will work adding a directed unit test.

mms/model_service_worker.py
```diff
@@ -177,8 +177,8 @@ def run_server(self):
         :return:
         """
         try:
-            self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
             self.sock.setblocking(True)  # workaround error(35, 'Resource temporarily unavailable') on OSX
+            self.sock.settimeout(SOCKET_ACCEPT_TIMEOUT)
```
frankfliu: This will crash on my mac

PR #520, opened 2018-09-12
Add models arguments to cli

frankfliu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
@@ -37,6 +37,11 @@ import java.security.GeneralSecurityException;
 import java.util.Collection;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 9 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
```diff
@@ -111,24 +129,25 @@ public void initModelStore() throws InvalidModelException, WorkerInitializationE
...
-                        if (!model.endsWith(".mar")) {
-                            modelFile = new File(modelStore, model + ".mar");
-                        }
+                    String[] pair = model.split("=", 2);
+                    String modelName = null;
+                    String url;
+                    if (pair.length == 1) {
+                        url = pair[0];
+                    } else {
+                        modelName = pair[0];
```
vrakesh: Should we validating urls here? in this file?

mms/model_server.py
```diff
@@ -2,31 +2,47 @@
...
 
 def start():
     """
     This is the entry point for model server
     :return:
     """
     args = ArgParser.mms_parser().parse_args()
-    if args.start is True:
-        os.environ['MODEL_SERVER_HOME'] = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    pid_file = tempfile.gettempdir() + "/.model_server.pid"
```
vrakesh: os.path.join() should be used instead

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
frankfliu: This is handled in ModelArchive

mms/model_server.py
frankfliu: Will change in following PR

PR #522, opened 2018-09-16
Enhance

frankfliu:
*Issue #, if available:*

*Description of changes:*


mms/service.py
```diff
@@ -11,15 +11,16 @@
...
 
 import mms
 from mms.context import Context
-from mms.log import log_msg
 from mms.metrics.metrics_store import MetricsStore
-from mms.mxnet_model_service_error import MMSError
-from mms.utils.validators.validate_messages import ModelWorkerMessageValidators
+from mms.protocol.otf_message_handler import create_predict_response
+
+INFERENCE_METRIC = 'InferenceTime'
```
vrakesh: RunTime will be a more accurate name here? Since we do not know the seperation between preprocess and actual inference?

mms/model_service/model_service.py
```diff
@@ -148,22 +139,9 @@ def inference(self, data):
...
-        post_start_ms = time.time()
         data = self._postprocess(data)
-        post_end_ms = time.time()
-
-        pre_time_in_ms = (infer_start_time - pre_start_time) * 1000
-        infer_time_in_ms = (post_start_ms - infer_start_time) * 1000
-        post_time_in_ms = (post_end_ms - post_start_ms) * 1000
-
-        metrics = self._context.metrics
-        metrics.add_time(PREPROCESS_METRIC, pre_time_in_ms)
```
vrakesh: why are the metrics being removed here?

mms/model_service_worker.py
```diff
@@ -10,70 +10,57 @@
...
-        self.send_failures = 0
-        self.codec = OtfCodecHandler()
-
-        try:
-            msg = "Listening on port: {}\n".format(s_name)
-            log_msg(msg)
-            socket_family = socket.AF_INET if s_type == 'tcp' else socket.AF_UNIX
-            self.sock = socket.socket(socket_family, socket.SOCK_STREAM)
-
-        except (IOError, OSError) as e:
```
vrakesh: Are we sure this exception need not be handled?

vrakesh:
Quite a few changes, since we do not have integration tests yet for the new version. Requesting some end-end manual tests, to verify functionality

mms/model_service_worker.py
frankfliu: If these exception is raised, the process will die. What do we need to handle?

mms/model_service/model_service.py
frankfliu: 1. Those are legacy models
2. They belong to model metrics. We don't have that part implemented yet.

mms/service.py
frankfliu: How about PredictionTime

vrakesh: That sounds good

PR #533, opened 2018-09-23
Enhance

frankfliu:
*Issue #, if available:*

*Description of changes:*


mms/model_server.py
```diff
@@ -19,20 +19,38 @@ def start():
...
     if args.stop is True:
-        if os.path.isfile(pid_file):
-            with open(pid_file, "r") as f:
-                try:
-                    os.kill(int(f.readline()), signal.SIGKILL)
-                except OSError:
-                    print("Model server already stopped")
-            os.remove(pid_file)
+        if pid is None:
+            print("Model server is not currently running.")
```
vrakesh: should this print be changed to logging

ci/buildspec.yml
```diff
@@ -33,7 +33,10 @@ phases:
       - $NIGHTLYBUILD
       - eval $NIGHTLYUPLOAD
 
-# local build will hang without artifacts section
 artifacts:
   files:
     - dist/*.whl
+    - model_archiver/dist/*.whl
+    - frontend/server/build/reports
+    - frontend/modelarchive/build/reports
```
vrakesh: reports are being added to nightly builds , why?

frankfliu: Report should be part of build artifacts. It's a good idea to keep all intermediate files for release.

mms/model_server.py
frankfliu: This file is only used for command line. print should be sufficient in this case.

docker/Dockerfile.gpu
```diff
@@ -16,10 +16,17 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server
+RUN pip install mxnet-model-server \
+    && pip install mxnet-cu90mkl
+
+COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
+RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
```
vdantu: Why not just run "mxnet-model-server" from the ENTRYPOINT?

docker/dockerd-entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+if [[ "$1" = "serve" ]]; then
+    shift 1
+    mxnet-model-server --start "$@"
+else
+    eval "$@"
```
vdantu: When do we get into this else?

frontend/cts/src/main/java/com/amazonaws/ml/mms/cts/Cts.java
```diff
@@ -69,14 +69,14 @@ private void startTest() {
                             "https://s3.amazonaws.com/mxnet-model-server/onnx-arcface/input2.jpg",
                             "player1.jpg");
 
-            HttpClient client = new HttpClient();
+            HttpClient client = new HttpClient(8081, 8080);
```
vdantu: What is this file? Its called Test but its under source. This should probably be moved to test folders if its a test file.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -162,13 +168,13 @@ public void setMemory(long memory) {
...
 
-    public void connect() throws WorkerInitializationException {
+    private void connect() throws WorkerInitializationException {
         if (!configManager.isDebug() && !lifeCycle.startWorker(port)) {
             throw new WorkerInitializationException(
                     ErrorCodes.INTERNAL_SERVER_ERROR_BACKEND_WORKER_INSTANTIATION);
         }
         String modelName = model.getModelName();
-        listener.notifyChangeState(modelName, WorkerStateListener.WORKER_STARTED);
+        setState(WorkerState.WORKER_STARTED);
```
vdantu: Why did we need to make these changes?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -140,20 +142,24 @@ public void run() {
             }
         } catch (InterruptedException e) {
             logger.warn("Backend worker thread interrupted.");
-        } catch (WorkerInitializationException wi) {
-            logger.error("Backend worker error", wi);
+        } catch (WorkerInitializationException e) {
```
vdantu: Why this change?

docker/Dockerfile.gpu
frankfliu: 1. use dockerd-entrypoint.sh is by convention
2. if use mxnet-model-server as ENTRYPOINT, it would be hard for user to override it, for example: "docker run -itd mms /bin/bash" won't work in that case.
3. ENTRYPOINT script should used for general purpose, and CMD is the parameter passed to the script.

docker/dockerd-entrypoint.sh
frankfliu: for example:
1. docker run -itd mms /bin/bash
2. docker run -itd mms mxnet-model-server --models ALL ...

frontend/cts/src/main/java/com/amazonaws/ml/mms/cts/Cts.java
frankfliu: This is not unittest and we don't want to run on every build.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
frankfliu: just make the exception handling more consistent.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
frankfliu: We can keep tracking the state in the worker.

frankfliu:
You can review by each commit. Each commit isn't big.

docker/dockerd-entrypoint.sh
vdantu: Oh. Looks like we would need to run the whole command anyways. Then we don't need entry-point.  What are we saving by adding this entry point?

frontend/cts/src/main/java/com/amazonaws/ml/mms/cts/Cts.java
vdantu: Oh. I am not sure what these files are or where they are used then. I will skip reviewing this.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
vdantu: Makes sense :)

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
vdantu: Ok. Makes sense. I was making similar changes, was curious why you were making this change or where you were using this.

docker/dockerd-entrypoint.sh
frankfliu: The one passed in is CMD, not entrypoint, to override entrypoint user has to use --entrypoint, which is not most user used to do.
The convention is use a fixed entrypoint named dockerd-entrypoint and use CMD as parameter. So it looks like you can run any commands, actually they just pass parameters to entrypoint.

vdantu: I see what's happening. I think you are using CMD as a default value to run the container. This looks good.

PR #534, opened 2018-09-25
Added ability to run all the benchmarks for a single model

piyushghai:
*Issue #, if available:*

*Description of changes:*
Improved scripts for Benchmark AI.

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -59,8 +59,8 @@ MODEL_NOOP = 'noop'
 MODEL_MAP = {
     MODEL_SQUEEZE_NET: (JMX_IMAGE_INPUT_MODEL_PLAN, {'url': 'https://s3.amazonaws.com/model-server/models/squeezenet_v1.1/squeezenet_v1.1.model', 'model_name': MODEL_SQUEEZE_NET, 'input_filepath': 'kitten.jpg'}),
     MODEL_RESNET_18: (JMX_IMAGE_INPUT_MODEL_PLAN, {'url': 'https://s3.amazonaws.com/model-server/models/resnet-18/resnet-18.model', 'model_name': MODEL_RESNET_18, 'input_filepath': 'kitten.jpg'}),
-    MODEL_LSTM_PTB: (JMX_TEXT_INPUT_MODEL_PLAN, {'url': 'https://s3.amazonaws.com/model-server/models/lstm_ptb/lstm_ptb.model', 'model_name': MODEL_LSTM_PTB}),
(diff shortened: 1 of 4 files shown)


benchmarks/lstm_ip.json
```diff
@@ -0,0 +1 @@
+"[{\"input_sentence\": \"on the exchange floor as soon as ual stopped trading we <unk> for a panic said one top floor trader\"}]"
```
zachgk: Missing newline

benchmarks/benchmark.py
```diff
@@ -472,7 +474,12 @@ def modify_config_props_for_mms(pargs):
 
     modify_config_props_for_mms(pargs)
 
-    if pargs.all:
+    if pargs.model:
```
zachgk: pargs.model has a default so this makes it impossible to reach the else case (run single benchmark on single model).  Add a new argument to the "target" mutually exclusive group for this additional case

benchmarks/benchmark.py
```diff
@@ -472,7 +474,12 @@ def modify_config_props_for_mms(pargs):
...
+        benchmark_model = pargs.model[0].lower()
+        for benchmark_name in ['latency', 'throughput']:
```
zachgk: Define the set of benchmarks as a constant for consistency

benchmarks/benchmark.py
piyushghai: Done.

benchmarks/benchmark.py
piyushghai: Done.

benchmarks/lstm_ip.json
piyushghai: As discussed offline, this is not needed here.

benchmarks/benchmark.py
```diff
@@ -444,6 +448,7 @@ def modify_config_props_for_mms(pargs):
     target = parser.add_mutually_exclusive_group(required=True)
     target.add_argument('name', nargs='?', type=str, choices=benchmark_name_options, help='The name of the benchmark to run')
     target.add_argument('-a', '--all', action='store_true', help='Run all benchmarks')
+    target.add_argument('--model-name', nargs=1, type=str, default=[MODEL_RESNET_18], choices=MODEL_MAP.keys(), help='A preloaded model to run.  It defaults to {}. Use this argument to run all the benchmarks on this single model'.format(MODEL_RESNET_18))
```
zachgk: Having this accept a model means there are two ways to pass in the model.  It is clearer to instead make this argument take a boolean similar to --all that simply changes the mode, and then reuse the --model/--custom-model parameter.  This better matches the behavior of name and all, as well as enables running the set of tests on custom models.  Maybe name running the set of tests for a given model "--suite"?

piyushghai: Good point. 
Done.

PR #536, opened 2018-09-25
Remove UDS sockets on worker shutdown

vrakesh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/mms/model_service_worker.py
+++ b/mms/model_service_worker.py
@@ -42,7 +42,7 @@ class MXNetModelServiceWorker(object):
                 raise ValueError("Wrong arguments passed. No socket name given.")
             self.sock_name, self.port = s_name, -1
             try:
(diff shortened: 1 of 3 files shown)


mms/model_service_worker.py
```diff
@@ -154,5 +154,8 @@ def run_server(self):
         logging.error("Backend worker did not receive connection in: %d", SOCKET_ACCEPT_TIMEOUT)
     except Exception:  # pylint: disable=broad-except
         logging.error("Backend worker process die.", exc_info=True)
+    finally:
+        if sock_type == 'unix' and os.path.exists(socket_name):
+            os.remove(socket_name)
```
frankfliu: This is not consistent with earlier one: os.unlink()
Can we change both to os.remove()?

PR #541, opened 2018-09-27
Implement new docker image spec.

frankfliu:
*Issue #, if available:
Implement new docker image spec.

*Description of changes:
1. set config.properties for docker image, override default parameters for docker image
2. /tmp folder might not accessible in some container, redirect to user home location
3. Allow user to configure JVM options that running in docker.
4. Allow user to configure model-store location via cli.

### DIFF
--- a/docker/Dockerfile.cpu
+++ b/docker/Dockerfile.cpu
@@ -16,16 +16,22 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server
+RUN pip install mxnet-model-server \
+    && pip install mxnet-mkl
+
+RUN useradd -m model-server \
+    && mkdir -p /home/model-server/tmp \
+    && chown -R model-server /home/model-server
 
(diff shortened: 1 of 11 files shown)


docker/Dockerfile.cpu
```diff
@@ -16,16 +16,22 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server
+RUN pip install mxnet-model-server \
+    && pip install mxnet-mkl
```
vdantu: This might need a rework once we release our mxnet-model-server. Do we need to add it now?

docker/config.properties
```diff
@@ -0,0 +1,14 @@
+# vmargs=-Xmx1g -XX:MaxDirectMemorySize=512m -Dlog4j.configuration=/opt/ml/conf/log4j.properties
+model_store=/opt/ml/model
+load_models=ALL
+inference_address=http://0.0.0.0:8080
+# management_address=http://127.0.0.1:8081
```
vdantu: By default we dont have management ports?

PR #544, opened 2018-09-28
Implement new docker image spec.

frankfliu:
1. set config.properties for docker image, override default parameters for docker image
2. /tmp folder might not accessible in some container, redirect to user home location
3. Allow user to configure JVM options that running in docker.
4. Allow user to configure model-store location via cli.
5. Improve initial model loading log message.


docker/Dockerfile.cpu
```diff
@@ -16,16 +16,22 @@ RUN apt-get update && \
...
+
 RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
 
-RUN useradd -m model-server
 USER model-server
 WORKDIR /home/model-server
-
-LABEL maintainer="dantu@amazon.com, rakvas@amazon.com, lufen@amazon.com, dden@amazon.com"
-
+ENV TEMP=/home/model-server/tmp
```
vdantu: Is this OK? Would sagemaker overwrite TEMP to some other directory? Or anyone else for that matter... Would we be better off creating our own temp instead of using something that is default, to avoid TEMP namespace collision?

vdantu: True. Maybe this is not a concern right away and we can address it later, but my concern is if we use these standard ones, we might get into issues when customers want to use TEMP/TMP or TMPDIR as something that they want to track ... For now I think its OK to tell customers that MMS requires read/write/modify access for TEMPDIR

PR #549, opened 2018-09-30
Fix backoff retry bug.

frankfliu:
Current implementation always retry in 0 seconds.

*Issue #, if available:*

*Description of changes:*


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -45,6 +45,10 @@
 
     static final Logger logger = LoggerFactory.getLogger(WorkerThread.class);
 
+    private static final int[] BACK_OFF = {
```
piyushghai: Can you rename this to Fibonacci back off or add a comment indicating the back off strategy is fibonacci ?

PR #556, opened 2018-10-01
Update installation document.

frankfliu:
*Issue #, if available:*

*Description of changes:*


docs/install.md
```diff
@@ -5,11 +5,41 @@
...
+pip install mxnet-cu90-mkl
+```
+
 
 * **Curl**: Optional. Curl is used in all of the examples. Install it with your preferred package manager.
 
-* **Unzip**: Optional. Unzip allows you to easily extract model files and inspect their content. If you choose to use it, associate it with `.model` extensions.
+* **Unzip**: Optional. Unzip allows you to easily extract model files and inspect their content. If you choose to use it, associate it with `.mar` extensions.
 
 ## Install MMS with pip
```
vrakesh: add mxnet installation option here as well?
eg: pip install mxnet-model-server[mxnet-mkl]

frankfliu: I didn't find a good way to add this. Will do another round of scrub with Aaron.

lupesko:
@aaronmarkham can you please help review the doc? Thanks!

docs/install.md
```diff
@@ -5,11 +5,41 @@
...
+```
+
+For Mac:
+```bash
+brew tap caskroom/versions
+brew update
+brew cask install java8
+```
+
+* **MxNet**: Recommended. MMS won't install `mxnet` by default. MxNet is required for most of examples in this project. MMS won't install mxnet engine by default, you can install mxnet-mkl or mxnet-cu90mkl based on your need. And you can also choose specific version of mxnet if you want.
```
vdantu: I think we do install. This needs to be rewritten to showcase the extras_require... Might be better to show the options. 
We should ask customers to do pip install if they have requirements for specific version of MXNet or other frameworks

PR #558, opened 2018-10-01
Added better logging for failed sync calls

vdantu:
*Issue #, if available:*
Made error messages more consistent to reflect current state of the system.
*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
@@ -238,28 +238,41 @@ public class ManagementRequestHandler extends HttpRequestHandler {
             int maxWorkers,
             boolean synchronous,
             final Function<Void, Void> onError) {
+
         ModelManager modelManager = ModelManager.getInstance();
         CompletableFuture<Boolean> future =
                 modelManager.updateModel(modelName, minWorkers, maxWorkers);
         if (!synchronous) {
             NettyUtils.sendJsonResponse(
-                    ctx, new StatusResponse("Worker updated"), HttpResponseStatus.ACCEPTED);
+                    ctx,
+                    new StatusResponse("Processing worker updates..."),
+                    HttpResponseStatus.ACCEPTED);
             return;
         }
         future.thenApply(
                         v -> {
+                            boolean status = modelManager.scaleRequestStatus(modelName);
+                            String response = "Workers scaled";
+                            HttpResponseStatus httpCode = HttpResponseStatus.OK;
+
                             if (!v) {
+                                response = "Workers scaling in progress";
                                 if (onError != null) {
                                     onError.apply(null);
+                                    response = "Load failed... Deregistered model " + modelName;
                                 }
                                 NettyUtils.sendError(
                                         ctx,
                                         HttpResponseStatus.NOT_FOUND,
-                                        new ModelNotFoundException(
-                                                "Model not found: " + modelName));
+                                        new ModelNotFoundException(response));
                             } else {
+                                if (!status) {
+                                    response = "Workers scaling in progress..";
+                                    httpCode = new HttpResponseStatus(210, "Partial Success");
+                                }
+
                                 NettyUtils.sendJsonResponse(
-                                        ctx, new StatusResponse("Worker scaled"));
+                                        ctx, new StatusResponse(response), httpCode);
                             }
                             return v;
                         })
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
@@ -68,14 +68,33 @@ public class WorkLoadManager {
         return worker.isEmpty();
     }
 
+    public boolean getWorkerStatus(String modelName) {
+        boolean status = true;
+        List<WorkerThread> threads = workers.getOrDefault(modelName, null);
+
+        if (threads == null) {
+            return true;
+        }
+
+        for (WorkerThread thread : threads) {
+            if ((thread.getState() == WorkerState.WORKER_STOPPED)
+                    || (thread.getState() == WorkerState.WORKER_ERROR)) {
+                status = false;
+                break;
+            }
+        }
+
+        return status;
+    }
+
     public CompletableFuture<Boolean> modelChanged(Model model) {
         synchronized (model.getModelName()) {
+            CompletableFuture<Boolean> future = new CompletableFuture<>();
             int minWorker = model.getMinWorkers();
             List<WorkerThread> threads;
             if (minWorker == 0) {
                 threads = workers.remove(model.getModelName());
                 if (threads == null) {
-                    CompletableFuture<Boolean> future = new CompletableFuture<>();
                     future.complete(Boolean.TRUE);
                     return future;
                 }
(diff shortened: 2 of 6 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/http/APIStatus.java
```diff
@@ -0,0 +1,7 @@
+package com.amazonaws.ml.mms.http;
+
+public enum APIStatus {
```
frankfliu: ApiStatus

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -86,19 +89,28 @@ public boolean hasNoWorker(String modelName) {
         if (currentWorkers < minWorker) {
             return addThreads(threads, model, minWorker - currentWorkers);
         } else {
-            for (int i = currentWorkers - 1; i >= minWorker; --i) {
-                WorkerThread thread = threads.remove(i);
-                thread.shutdown();
+            for (int i = 0; i < currentWorkers - 1; i++) {
```
frankfliu: You have to iterate backward. otherwise, threads.remove(i) will failed with ArrayIndexOutofBoundException, or ConcurrentModificationException

frankfliu: Should we do two pass scan?
1st scan for all workers in bad state, and then blindly kill workers.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -86,19 +89,28 @@ public boolean hasNoWorker(String modelName) {
...
+                if (i >= minWorker) {
+                    WorkerThread thread = threads.remove(i);
+                    thread.shutdown();
+                } else {
+                    if (threads.get(i).getState() == WorkerState.WORKER_ERROR
+                            || threads.get(i).getState() == WorkerState.WORKER_STOPPED) {
+                        status = APIStatus.PARTIAL_SUCCESS;
```
frankfliu: Should PARTIAL_SUCCESS only apply to scale up?

vdantu: I think this should apply to any scale call. For example, this should also apply for two consecutive calls for same number of workers.

frankfliu: This behavior is not consistent with scale up.
When scale down, if any existing worker is failed, we return partial success, but scale up, we only check added worker.

The simplest behavior is return partial if:
1. Success = all worker are healthy
2. At least worker is success, but have some worker failed
3. Failed = all worker are failed

We actually don't need any of these changes. We can simply loop all workers to check health status before API return.

vdantu: True, its simpler to hold all the changes in the same method.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -239,16 +239,18 @@ private void updateModelWorkers(
...
-                    ctx, new StatusResponse("Worker updated"), HttpResponseStatus.ACCEPTED);
+                    ctx,
+                    new StatusResponse("Processing Workers scale request..."),
+                    HttpResponseStatus.ACCEPTED);
             return;
         }
         future.thenApply(
                         v -> {
-                            if (!v) {
+                            if (v == ApiStatus.FAIL) {
```
frankfliu: We don't really care about ApiStatus returned by future, only thing we need to do is loop through workers, and check current status of each workers.

Even CompletableFuture<Void> should be enough.
As I mentioned, this is the only place need change. All other changes are not necessary.

frankfliu: I would like to keep it simple for now.
If we don't have a ideal solution, leave it for future.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -68,14 +68,33 @@ public boolean hasNoWorker(String modelName) {
         return worker.isEmpty();
     }
 
+    public boolean getWorkerStatus(String modelName) {
+        boolean status = true;
```
frankfliu: This local variable is not necessary. You can return ture/false in the loop.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -68,14 +68,33 @@ public boolean hasNoWorker(String modelName) {
...
+                break;
+            }
+        }
+
+        return status;
+    }
+
     public CompletableFuture<Boolean> modelChanged(Model model) {
         synchronized (model.getModelName()) {
+            CompletableFuture<Boolean> future = new CompletableFuture<>();
```
frankfliu: Thanks for making this change.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -243,23 +243,30 @@ private void updateModelWorkers(
...
-                                        new ModelNotFoundException(
-                                                "Model not found: " + modelName));
+                                        new ModelNotFoundException(response));
                             } else {
-                                NettyUtils.sendJsonResponse(
-                                        ctx, new StatusResponse("Worker scaled"));
+                                if (!status) {
+                                    response = "Workers scaling in progress..";
+                                }
+                                NettyUtils.sendJsonResponse(ctx, new StatusResponse(response));
```
frankfliu: We should send non 200 code back if status is false.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
vdantu: I would like to keep it this way. I prefer having as little exits from a method as possible.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
vdantu: OH. I thought it was a success since we already created the workers and we are just retrying to load again and again. What code would you suggest best reflects this? Should we send a 5XX error for sync-calls?

frankfliu: How about 206?

vdantu: But 206 is partial content.. https://tools.ietf.org/html/rfc7233#section-4.1 .. RFC suggests that its for returning partial content... Quote from the RFC
```
   If a single part is being transferred, the server generating the 206
   response MUST generate a Content-Range header field, describing what
   range of the selected representation is enclosed, and a payload
   consisting of the range
```

Not sure if this would affect client behaviors...

frankfliu: Agree, I think we can make up one: 210 partial success.

vdantu: Yup. Added that.

PR #564, opened 2018-10-02
Update API to match document.

frankfliu:
*Issue #, if available:*

*Description of changes:*


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -87,7 +88,8 @@ public boolean hasNoWorker(String modelName) {
             if (currentWorkers < minWorker) {
                 return addThreads(threads, model, minWorker - currentWorkers);
             } else {
-                for (int i = currentWorkers - 1; i >= minWorker; --i) {
+                for (int i = currentWorkers - 1; i >= maxWorker; --i) {
```
vdantu: Whats this change? Why is maxWorker being used instead of minWorker?

frankfliu: We should respect input parameter. In current implementation:
1. when scale up, we scale to minWorkers.
2. when sale down, we scale to maxWorkers.
3. We don't do auto adjust between min - max.

This behavior matches user expectation. We we don't need run into future compatibility issue.

vdantu: Ok. I am going to reluctantly give an approval here. 
This API seems confusing because each scale call requires the customers to give two parameters now. As I see it, customers are only going to give the same value for both parameters. I think we are getting into realm of needing to explain to customers how to use this API even more. IMHO, using min_workers, or even better removing both and having only num_workers seems simpler to maintain for this release and probably even coming few releases.

frankfliu: No, none of them are not required.
if max is not specified, max is the same as min.

PR #567, opened 2018-10-04
Fix Ping bug

vdantu:
*Issue #, if available:*
Fixed a bug with ping. The behavior needed to change after the backoff timer changes.

*Description of changes:*


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -162,24 +163,33 @@ public void workerStatus(final ChannelHandlerContext ctx) {
...
         wlm.scheduleAsync(r);
     }
 
     public boolean scaleRequestStatus(String modelName) {
-        return wlm.getWorkerStatus(modelName);
+        Model model = ModelManager.getInstance().getModels().get(modelName);
+        int numWorkers = wlm.getNumRunningWorkers(modelName);
+        boolean success = true;
+
+        if ((model != null) && (model.getMinWorkers() > numWorkers)) {
```
frankfliu: this can be simplified as:
return model == null || model.getMinWorkers() <= numWorkers;

PR #569, opened 2018-10-05
Python workers are sent all the environment variables which aren't in the blacklist

vdantu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
@@ -36,8 +36,11 @@ import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Enumeration;
+import java.util.InvalidPropertiesFormatException;
 import java.util.List;
 import java.util.Properties;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
@@ -55,6 +58,7 @@ public final class ConfigManager {
     private static final String MODEL_SERVER_HOME = "model_server_home";
     private static final String MODEL_STORE = "model_store";
     private static final String LOAD_MODELS = "load_models";
+    private static final String BLACKLIST_ENV_VARS = "blacklist_env_vars";
     private static final String NUMBER_OF_NETTY_THREADS = "number_of_netty_threads";
     private static final String MAX_WORKERS = "max_workers";
     private static final String JOB_QUEUE_SIZE = "job_queue_size";
@@ -67,7 +71,13 @@ public final class ConfigManager {
     private static final String CERTIFICATE_FILE = "certificate_file";
     private static final String PRIVATE_KEY_FILE = "private_key_file";
 
+    private Pattern blacklistPattern;
+
     private Properties prop;
+
+    public Pattern getBlacklistPattern() {
+        return blacklistPattern;
+    }
 
     public ConfigManager(Arguments args) {
         prop = new Properties();
@@ -288,6 +298,15 @@ public final class ConfigManager {
         return prop.getProperty(key, def);
     }
 
+    public void validateConfigurations() throws InvalidPropertiesFormatException {
+        String blacklistVars = prop.getProperty(BLACKLIST_ENV_VARS, "");
+        try {
+            blacklistPattern = Pattern.compile(blacklistVars);
+        } catch (PatternSyntaxException e) {
+            throw new InvalidPropertiesFormatException(e);
+        }
+    }
+
     public String dumpConfigurations() {
         return "\nMMS Home: "
                 + getModelServerHome()
@@ -304,7 +323,9 @@ public final class ConfigManager {
                 + "\nLog dir: "
                 + getCanonicalPath(System.getProperty("LOG_LOCATION"))
                 + "\nMetrics dir: "
-                + getCanonicalPath(System.getProperty("METRICS_LOCATION"));
+                + getCanonicalPath(System.getProperty("METRICS_LOCATION"))
+                + "\nBlacklist Regex: "
+                + prop.getProperty(BLACKLIST_ENV_VARS, "");
     }
 
     void setProperty(String key, String value) {
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
@@ -37,9 +37,6 @@ public class Model {
 
     public Model(ModelArchive modelArchive, int queueSize) {
         this.modelArchive = modelArchive;
-        minWorkers = 1;
-        maxWorkers = 1;
-
         batchSize = 1;
         maxBatchDelay = 100;
         jobsDb = new ConcurrentHashMap<>();
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
@@ -21,9 +21,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.SocketAddress;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Scanner;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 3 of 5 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -69,6 +71,11 @@
 
     private Properties prop;
 
+    public HashSet<String> getBlacklistEnvVars() {
+        return new HashSet<String>(
+                Arrays.asList(prop.getProperty(BLACKLIST_ENV_VARS).split("\\s*,\\s*")));
```
frankfliu: You might get an NPE

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -37,8 +37,8 @@
 
     public Model(ModelArchive modelArchive, int queueSize) {
         this.modelArchive = modelArchive;
-        minWorkers = 1;
-        maxWorkers = 1;
+        minWorkers = 0;
```
frankfliu: Just remove them

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,20 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
         this.model = model;
     }
 
+    ArrayList<String> getEnvString(HashMap<String, String> envMap) {
+        ArrayList<String> envList = new ArrayList<>();
+        HashSet<String> blackList = configManager.getBlacklistEnvVars();
+
+        if (!blackList.contains("ALL")) {
```
frankfliu: if contains all, returning empyt list?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,20 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
...
+            for (Map.Entry<String, String> entry : envMap.entrySet()) {
+                if (!blackList.contains(entry.getKey()) || blackList.contains("NONE")) {
```
frankfliu: A empty blacklist means none, A special "NONE" isn't really meaningful.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,20 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
...
+                    envList.add(entry.getKey() + "=" + entry.getValue());
```
frankfliu: '='

frontend/server/src/test/resources/config.properties
```diff
@@ -4,6 +4,7 @@ management_address=https://127.0.0.1:8444
 model_server_home=../..
 model_store=../modelarchive/src/test/resources/models
 load_models=noop-v0.1,noop-v1.0
+blacklist_env_vars = NONE
```
frankfliu: This property should be optional

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
vdantu: I agree. Probably I should remove all because ALL can potentially be a valid environment variable. Should I replace it with MMS_ALL... I will remove all for now.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -69,6 +71,17 @@
 
     private Properties prop;
 
+    public HashSet<String> getBlacklistEnvVars() {
+        HashSet<String> res = new HashSet<>();
```
frankfliu: if (blacklistVars == null) {
return Collections.emptySet();
}
return new HashSet<String>(Arrays.asList(...));

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,29 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
         this.model = model;
     }
 
+    private String[] getEnvString(String cwd, String modelPath) {
+        ArrayList<String> envList = new ArrayList<>();
+        HashSet<String> blackList = configManager.getBlacklistEnvVars();
```
frankfliu: I think use a single regex string for blackList might be better then delimited set

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,35 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
         this.model = model;
     }
 
+    private String[] getEnvString(String cwd, String modelPath) {
+        ArrayList<String> envList = new ArrayList<>();
+        String blackList = "";
+        try {
+            blackList = configManager.getBlacklistEnvVars();
+        } catch (PatternSyntaxException e) {
```
frankfliu: If the regex is checked at startup time, we don't need catch this exception here.

frontend/server/src/test/resources/config.properties
```diff
@@ -14,3 +14,4 @@ load_models=noop-v0.1,noop-v1.0
 # keystore_type=PKCS12
 private_key_file=src/test/resources/key.pem
 certificate_file=src/test/resources/certs.pem
+# blacklist_egit nv_vars=.*USERNAME.*||.*PASSWORD.*
```
frankfliu: nv_vars=("|')?(AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?("|')?\s*(:|=>|=)\s*("|')?[0-9]{4}\-?[0-9]{4}\-?[0-9]{4}("|')?

vdantu: What is this comment for?

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -69,6 +72,17 @@
 
     private Properties prop;
 
+    public String getBlacklistEnvVars() throws PatternSyntaxException {
```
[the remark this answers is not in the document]
vdantu: The check can be done at start time, true. Will change it.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,35 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
...
+        if (System.getenv("PYTHONPATH") != null) {
+            pythonPath.append(System.getenv("PYTHONPATH")).append(File.pathSeparatorChar);
+        }
+
+        pythonPath.append(modelPath).append(File.pathSeparatorChar).append(cwd);
+
+        environment.put("PYTHONPATH", pythonPath.toString());
+
+        for (Map.Entry<String, String> entry : environment.entrySet()) {
+            if (!entry.getKey().matches(blackList)) {
```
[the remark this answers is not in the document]
vdantu: It isn't compiled each time. We compile it only once per worker creation.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -69,6 +72,17 @@
...
+        String res = "";
+        String blacklistVars = prop.getProperty(BLACKLIST_ENV_VARS, null);
+
+        if (blacklistVars != null) {
+            Pattern.compile(blacklistVars);
```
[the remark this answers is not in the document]
vdantu: We compile it once per worker. Shouldn't be a huge hit, but I moved it to compile it at start-up.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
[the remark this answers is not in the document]
frankfliu: String.matches(regex) method internal compile the regex on each call.

frontend/server/src/test/resources/config.properties
frankfliu: This is a real example of blacklist AWS secrets.
And the example you profile isn't correct, it should | not be ||

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -288,7 +297,18 @@ public String getProperty(String key, String def) {
         return prop.getProperty(key, def);
     }
 
+    public void validateConfigurations() {
+        String blacklistVars = prop.getProperty(BLACKLIST_ENV_VARS, "");
+        try {
+            blacklistPattern = Pattern.compile(blacklistVars);
+        } catch (PatternSyntaxException e) {
+            prop.setProperty(BLACKLIST_ENV_VARS, "");
```
frankfliu: print warning here.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -45,6 +49,30 @@ public WorkerLifeCycle(ConfigManager configManager, Model model) {
         this.model = model;
     }
 
+    private String[] getEnvString(String cwd, String modelPath) {
+        ArrayList<String> envList = new ArrayList<>();
```
frankfliu: In java convention, variable declaration should be as closer to first usage as possible.

PR #573, opened 2018-10-08
Gluon alexnet example changes

vdantu:
*Issue #, if available:*
Gluon Alexnet example
*Description of changes:*
TODO:
I will go through README and update if required.

### DIFF
--- a/examples/gluon_alexnet/README.md
+++ b/examples/gluon_alexnet/README.md
@@ -15,9 +15,7 @@ This tutorial covers the following:
 time, but are familiar with creating a neural network with MXNet or another framework, you may refer this 10 min Gluon crash-course: [Predict with a pre-trained model](http://gluon-crash-course.mxnet.io/predict.html).
 * **Gluon naming**. Fine-tuning pre-trained Gluon models requires some understanding of how the naming conventions work. Take a look at the [Naming of Gluon Parameter and Blocks](https://mxnet.incubator.apache.org/tutorials/gluon/naming.html) tutorial for more information.
 * **Basic MMS knowledge**. If you are using MMS for the first time, you should take advantage of the [MMS QuickStart tutorial](https://github.com/awslabs/mxnet-model-server#quick-start).
-* **MMS Custom Service knowledge.** Review the [Defining a Custom Service](https://github.com/awslabs/mxnet-model-server/blob/master/docs/custom_service.md) documentation to understand how MMS implements custom pre and post-processing for models.
 * **MMS installed**. If you haven't already, [install MMS with pip](https://github.com/awslabs/mxnet-model-server/blob/master/docs/install.md#install-mms-with-pip) or [install MMS from source](https://github.com/awslabs/mxnet-model-server/blob/master/docs/install.md#install-mms-from-source-code). Either installation will also install MXNet.
-
 
 Refer to the [MXNet model zoo](https://mxnet.incubator.apache.org/api/python/gluon/model_zoo.html) documentation for examples of accessing other models.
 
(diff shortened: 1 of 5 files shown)


examples/template/gluon_base_service.py
```diff
@@ -0,0 +1,130 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""`Gluon vision service` defines a Gluon base vision service
```
frankfliu: Gluon base service

examples/template/gluon_base_service.py
```diff
@@ -0,0 +1,130 @@
...
+"""
+import mxnet as mx
+import numpy as np
+import os
+import json
+from mms.utils.mxnet import ndarray
+
+
+class GluonBaseService(object):
+    """MXNetVisionService defines a fundamental service for image classification task.
```
frankfliu: GluonBaseService

frankfliu: Start comments on 2nd line

examples/gluon_alexnet/gluon_hybrid_alexnet.py
```diff
@@ -10,7 +10,7 @@
 
 from mxnet.gluon import nn
 from mxnet.gluon.block import HybridBlock
-from mms.model_service.gluon_vision_service import GluonVisionService
+from ..template.gluon_base_service import GluonBaseService
```
frankfliu: template won't be part file path in model archive.

We can assume gluon_base_service.py directly in the model directory.

examples/gluon_alexnet/README.md
```diff
@@ -15,8 +15,13 @@ This tutorial covers the following:
...
 * **Gluon naming**. Fine-tuning pre-trained Gluon models requires some understanding of how the naming conventions work. Take a look at the [Naming of Gluon Parameter and Blocks](https://mxnet.incubator.apache.org/tutorials/gluon/naming.html) tutorial for more information.
 * **Basic MMS knowledge**. If you are using MMS for the first time, you should take advantage of the [MMS QuickStart tutorial](https://github.com/awslabs/mxnet-model-server#quick-start).
-* **MMS Custom Service knowledge.** Review the [Defining a Custom Service](https://github.com/awslabs/mxnet-model-server/blob/master/docs/custom_service.md) documentation to understand how MMS implements custom pre and post-processing for models.
 * **MMS installed**. If you haven't already, [install MMS with pip](https://github.com/awslabs/mxnet-model-server/blob/master/docs/install.md#install-mms-with-pip) or [install MMS from source](https://github.com/awslabs/mxnet-model-server/blob/master/docs/install.md#install-mms-from-source-code). Either installation will also install MXNet.
+* **MMS Configuration**. Add the following configuration to "/tmp" directory as "/tmp/mms_config"
```
frankfliu: We don't really need this configuration.

examples/gluon_alexnet/README.md
```diff
@@ -71,13 +91,13 @@ cp gluon_pretrained_alexnet.py synset.txt signature.json /tmp/models
...
-mxnet-model-export --model-name="pretrained-alexnet" --model-path="/tmp/models" --service-file-path="/tmp/models/gluon_pretrained_alexnet.py"
+model-archiver --model-name alexnet --model-path /tmp/models --handler gluon_pretrained_alexnet:pretrained_gluon_alexnet --runtime python --export-path /tmp
 ```
-This creates a model-archive file `pretrained-alexnet.model`.
+This creates a model-archive file `/tmp/alexnet.mar`.
 
 4. You could run the server with this model file to serve the pre-trained alexnet.
 ```bash
-mxnet-model-server --models alexnet=pretrained-alexnet.model
+mxnet-model-server --start --models /tmp/alexnet.mar --mms-config /tmp/mms_config
```
frankfliu: mxnet-model-server --start --models alexnet.mar --model-store /tmp

So you don't need a config file.

examples/gluon_alexnet/gluon_hybrid_alexnet.py
vdantu: It should be a part of this file, otherwise there would be errors in this file, albeit its a example code.

examples/gluon_alexnet/README.md
frankfliu: configuration file is for advanced user. We should keep the example as simple as possible.

examples/gluon_alexnet/gluon_hybrid_alexnet.py
frankfliu: User will copy files under template to /tmp/crepe, not the template folder
The files in /tmp/crepe will be: /tmp/crepe/gluon_vision.service.py,
So the import should be:

from .gluon_base_service import GluonBaseService

vdantu: True. That's how we test it, but having .gluon_base_service would mislead people going through this code because that file doesn't exist in this folder. People reading python code would understand that this service file is in the ```template``` folder. Otherwise, the code is not correct. If we want a working example, we should have proper references to the module. This is similar to the base service code using mxnet_utils. The users could have all the util files also in the same place.

PR #577, opened 2018-10-09
Created new alexnet model and modified the model_zoo document

vdantu:
*Issue #, if available:*
Created a new alexnet model archive file and updated the model zoo link

*Description of changes:*

### DIFF
--- a/docs/model_zoo.md
+++ b/docs/model_zoo.md
@@ -8,7 +8,7 @@ To propose a model for inclusion, please submit a [pull request](https://github.
 
 | Model File | Type | Dataset | Source | Size | Download |
 | --- | --- | --- | --- | --- | --- |
-| [AlexNet](#alexnet) | Image Classification | ImageNet | ONNX | 233 MB | [.mar](https://s3.amazonaws.com/model-server/model_archive_1.0/onnx-alexnet/alexnet.mar) |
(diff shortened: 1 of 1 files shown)


docs/model_zoo.md
```diff
@@ -8,7 +8,7 @@ To propose a model for inclusion, please submit a [pull request](https://github.
 
 | Model File | Type | Dataset | Source | Size | Download |
 | --- | --- | --- | --- | --- | --- |
-| [AlexNet](#alexnet) | Image Classification | ImageNet | ONNX | 233 MB | [.mar](https://s3.amazonaws.com/model-server/model_archive_1.0/onnx-alexnet/alexnet.mar) |
+| [AlexNet](#alexnet) | Image Classification | ImageNet | ONNX | 233 MB | [.mar](https://s3.amazonaws.com/model-server/models/alexnet/alexnet.mar) |
```
frankfliu: We should put all 1.0 model archive file in the same location on S3.

vdantu: I put it where our other models are. There are other models in this bucket as well.

vdantu: OH, I think you created a new bucket for 1.0 archive. Not sure why you created a new bucket, they are all just models, we could have let them all live in the same bucket. But, anyways, I changed the location.

PR #581, opened 2018-10-10
Detect number gpu on MMS startup.

frankfliu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
@@ -111,6 +111,10 @@ public class ModelServer {
         }
 
         ModelManager modelManager = ModelManager.getInstance();
+        int workers = configManager.getNumberOfGpu();
+        if (workers == 0) {
+            workers = Runtime.getRuntime().availableProcessors();
+        }
         if ("ALL".equalsIgnoreCase(loadModels)) {
             String modelStore = configManager.getModelStore();
             if (modelStore == null) {
@@ -141,7 +145,7 @@ public class ModelServer {
                         logger.debug("Loading models from model store: {}", file.getName());
 
                         ModelArchive archive = modelManager.registerModel(file.getName());
-                        modelManager.updateModel(archive.getModelName(), 1, 1);
+                        modelManager.updateModel(archive.getModelName(), workers, workers);
                     } catch (ModelException | IOException e) {
                         logger.warn("Failed to load model: " + file.getAbsolutePath(), e);
                     }
@@ -170,7 +174,7 @@ public class ModelServer {
 
                 ModelArchive archive =
                         modelManager.registerModel(url, modelName, null, null, 1, 100);
-                modelManager.updateModel(archive.getModelName(), 1, 1);
+                modelManager.updateModel(archive.getModelName(), workers, workers);
             } catch (ModelException | IOException e) {
                 logger.warn("Failed to load model: " + url, e);
             }
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
@@ -122,6 +122,10 @@ public final class ConfigManager {
         if (models != null) {
             prop.setProperty(LOAD_MODELS, String.join(",", models));
         }
+
+        if (!prop.containsKey(NUMBER_OF_GPU)) {
+            prop.setProperty(NUMBER_OF_GPU, String.valueOf(getAvailableGpu()));
+        }
     }
 
     public boolean isDebug() {
@@ -316,6 +320,8 @@ public final class ConfigManager {
                 + System.getProperty("java.io.tmpdir")
                 + "\nConfig file: "
                 + prop.getProperty("mmsConfigFile")
+                + "\nNumber of GPU: "
+                + getNumberOfGpu()
                 + "\nModel Store: "
                 + getModelStore()
                 + "\nInitial Models: "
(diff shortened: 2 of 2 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -368,6 +374,24 @@ private static String getCanonicalPath(String path) {
         return getCanonicalPath(new File(path));
     }
 
+    private static int getAvailableGpu() {
+        try {
+            Process process =
+                    Runtime.getRuntime().exec("nvidia-smi --query-gpu=index --format=csv");
```
piyushghai: This will only work for Linux machines. 
Might not work for other OS Types.

frankfliu: Why you think this won't work on other OS?

piyushghai: An example : 
https://devtalk.nvidia.com/default/topic/839954/cuda-programming-and-performance/command-line-gpu-load-display-on-os-x/post/4563535/#4563535

piyushghai: From a more official source : https://developer.nvidia.com/nvidia-system-management-interface

```
NVIDIA-smi ships with NVIDIA GPU display drivers on Linux, and with 64bit Windows Server 2008 R2 and Windows 7. Nvidia-smi can report query information as XML or human readable plain text to either standard output or a file. For more details, please refer to the nvidia-smi documentation.
```

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -122,6 +122,10 @@ public ConfigManager(Arguments args) {
         if (models != null) {
             prop.setProperty(LOAD_MODELS, String.join(",", models));
         }
+
+        if (!prop.containsKey(NUMBER_OF_GPU)) {
```
vdantu: We shouldn't use GPU if the properties doesn't contain GPU key. Customers might not want to use GPUs.

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
```diff
@@ -111,6 +111,10 @@ private void initModelStore() {
         }
 
         ModelManager modelManager = ModelManager.getInstance();
+        int workers = configManager.getNumberOfGpu();
```
vdantu: Why are workers tied to number of GPUs? They should be independent.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: There should be a special value (-1) or some such, which should mean auto-detect GPUs.

frankfliu: 1. number_of_gpu = 0 is still a valid configuration.
2. If user didn't config gpu, auto detect should be the default. What's the use case we don't want auto detect as default?
3. We don't need special value (-1), user didn't add this properties in config file is the indicator of auto detection.

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
frankfliu: What should be default number of worker when server startup?
Isn't this the right behavior?

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: 1. It should be valid.
2. Auto-detection shouldn't happen if customers don't want to use num GPUs option. We don't understand the customer use case. They might be running training on the same host and GPUs might be used for that and they might want to run inference only on CPUs. So, we should make it explicit to use GPUs and not auto-detect. 
3. If we go by point 2, then we would need a value for auto-detect. The value could be whatever we choose.
4. We should do auto-detect for any value larger than 0. The actual value should be [0,  min(given_number_of_gpu, total_available_gpu)] if a value is given.

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
vdantu: We should let MMS users configure this. If they want 1 worker we create only 1 worker.  Falling back to num-gpu's should be used as only if customers want us to fallback (customers should be given an option to choose this as well). The intuition for this is, num-gpu number of workers isn't the most efficient number of workers for most cases. If anything, this should be num-vcpus number of workers, but that might fail if we try to load all these workers models onto GPUs. Example of this could be if a host has num-vcpus is 64 and num_gpus is 1 with 8GB and each model worker requires 1GB of memory. Since there is no magic bullet solution yet to pick the right number of workers, we should let customers decide with a fallback option to num_gpus or num_cpus.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: @piyushghai : Do you know of any tool which can be used across platforms?

piyushghai: None that I'm aware of.

ddavydenko:
I suggest we move ahead with the solution of autodetecting GPUs at startup time and spawning amount of workers equal to GPUs ***if*** current instance host is GPU. Seems like that is what the code is doing by handling exception if there are no GPUs found and falling back to CPU count for workers to start with.

PR #589, opened 2018-10-12
Modified error messages thrown from Model Archiver

vdantu:
*Issue #, if available:*

*Description of changes:*


model-archiver/model_archiver/model_packaging.py
```diff
@@ -38,7 +38,9 @@ def package_model(args, manifest):
         # Step 3 : Zip 'em all up
         ModelExportUtils.zip(export_file_path, model_name, model_path, files_to_exclude, manifest)
         logging.info("Successfully exported model %s to file %s", model_name, export_file_path)
-
+    except Exception as e:  # pylint: disable=broad-except
+        logging.error(e)
```
frankfliu: For ModelArchiverError, we should not print stacktrace, but other exception, we should.

model-archiver/model_archiver/model_packaging_utils.py
```diff
@@ -57,11 +57,10 @@ def check_mar_already_exists(model_name, export_file_path, overwrite):
             if overwrite:
                 logging.warning("Overwriting %s ...", export_file)
             else:
-                logging.error("%s already exists.\n"
-                              "Please specify --force/-f option to overwrite the model archive output file.\n"
-                              "See -h/--help for more details.", export_file)
-
-                sys.exit(1)
+                raise ModelArchiverError("%s already exists.\n"
+                                         "Please specify --force/-f option to overwrite the model archive "
```
frankfliu: export_file_path is expected to be folder, We should check export_file_path exists and it's has to be a folder.

model-archiver/model_archiver/model_packaging.py
```diff
@@ -38,7 +38,9 @@ def package_model(args, manifest):
         # Step 3 : Zip 'em all up
         ModelExportUtils.zip(export_file_path, model_name, model_path, files_to_exclude, manifest)
         logging.info("Successfully exported model %s to file %s", model_name, export_file_path)
-
+    except Exception as e:  # pylint: disable=broad-except
+        logging.exception(e)
```
frankfliu: For ModelArchiverError we should not print stacktrace, but other error we should keep stacktrace.

model-archiver/model_archiver/model_packaging_utils.py
vdantu: This is already handled. You get "Not a directory:" error

model-archiver/model_archiver/model_packaging.py
vdantu: Fixed.

model-archiver/model_archiver/model_packaging.py
```diff
@@ -38,7 +39,12 @@ def package_model(args, manifest):
         # Step 3 : Zip 'em all up
         ModelExportUtils.zip(export_file_path, model_name, model_path, files_to_exclude, manifest)
         logging.info("Successfully exported model %s to file %s", model_name, export_file_path)
-
+    except Exception as e:  # pylint: disable=broad-except
```
frankfliu: except ModelArchiverError as e:
    logging.errror(e)
    sys.exit(1)

We don't need catch Exception, it will be printed with stacktrace.

model-archiver/model_archiver/model_packaging_utils.py
vdantu:
```
$ model-archiver --model-name alexnet --model-path onnx_alexnet/ --export-path /tmp/alexnet.mar -f --handler mxnet_vision_service:handle
ERROR:root:Failed to save the model-archive to model-path "/tmp/alexnet.mar". Check the file permissions and retry.
ERROR:root:[Errno 20] Not a directory: '/tmp/alexnet.mar/alexnet.mar'
Traceback (most recent call last):
  File "/Users/dantu/PycharmProjects/mxnet-model-server/model-archiver/model_archiver/model_packaging.py", line 40, in package_model
    ModelExportUtils.zip(export_file_path, model_name, model_path, files_to_exclude, manifest)
  File "/Users/dantu/PycharmProjects/mxnet-model-server/model-archiver/model_archiver/model_packaging_utils.py", line 232, in zip
    with zipfile.ZipFile(mar_path, 'w', zipfile.ZIP_DEFLATED) as z:
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 1090, in __init__
    self.fp = io.open(file, filemode)
NotADirectoryError: [Errno 20] Not a directory: '/tmp/alexnet.mar/alexnet.mar'
```
I might be missing something, would like to understand what it is. Is the suggestion to make sure that we catch "not a directory" error or to modify the error string?

model-archiver/model_archiver/model_packaging.py
vdantu: True, I used logging.exception() to print general exceptions with stacktrace. I will work on modifying this, but please do let me know if there is anything wrong with using logging.exception().

model-archiver/model_archiver/model_packaging_utils.py
frankfliu: We should not print stacktrace for input error cases. Instead we should give clear error message.

And the format of message should not be normally logging format. We should provide a good user experience as a normal command line tool.

vdantu: Ok. I am guessing you are asking to validate the input params. We could add that.

model-archiver/model_archiver/model_packaging_utils.py
```diff
@@ -213,11 +218,28 @@ def clean_temp_files(temp_files):
...
-            z.writestr(os.path.join(MAR_INF, MANIFEST_FILE_NAME), manifest)
+        try:
+            with zipfile.ZipFile(mar_path, 'w', zipfile.ZIP_DEFLATED) as z:
+                ModelExportUtils.zip_dir(model_path, z, set(files_to_exclude))
+                # Write the manifest here now as a json
+                z.writestr(os.path.join(MAR_INF, MANIFEST_FILE_NAME), manifest)
+        except IOError:
+            logging.error("Failed to save the model-archive to model-path \"%s\". "
+                          "Check the file permissions and retry.", export_file)
+            raise
```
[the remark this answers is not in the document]
vdantu: I am leaving regular Exceptions as is. That keeps the original Exception/Error stacktrace and message. Using the model archiver errors for the errors that we generate.

model-archiver/model_archiver/model_packaging_utils.py
```diff
@@ -125,22 +118,25 @@ def convert_onnx_model(model_path, onnx_file, model_name):
...
         symbol_file = '%s-symbol.json' % model_name
         params_file = '%s-0000.params' % model_name
         signature_file = 'signature.json'
         # Find input symbol name and shape
-        model_proto = onnx.load(os.path.join(model_path, onnx_file))
+        try:
+            model_proto = onnx.load(os.path.join(model_path, onnx_file))
+        except:
+            logging.error("Failed to load the %s model. Verify if the model file is valid", onnx_file)
+            raise
```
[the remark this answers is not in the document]
vdantu: Same as above.

model-archiver/model_archiver/model_packaging_utils.py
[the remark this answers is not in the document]
frankfliu: This is not an expected error cases, we should not convert to ModelArchiverException

model-archiver/model_archiver/model_packaging.py
```diff
@@ -38,7 +40,9 @@ def package_model(args, manifest):
         # Step 3 : Zip 'em all up
         ModelExportUtils.zip(export_file_path, model_name, model_path, files_to_exclude, manifest)
         logging.info("Successfully exported model %s to file %s", model_name, export_file_path)
-
+    except ModelArchiverError as e:
+        logging.error(e)
```
frankfliu: The error message looks like following:
ERROR:root:Given export-path a is not a directory. Point to a valid export-path directory.

Can we configure logging to only print out the message?

vdantu: Sure. We could make that. I don't think this is a blocker for PR though.

PR #609, opened 2018-10-16
Use python which MMS installed with to start metrics collector and mo…

frankfliu:
MetricsCollector depends psutil, which installed together with MMS.
MMS java code should use the same python to start metrics collector.

When start model service code, if generic python runtime is specified (default), we should try to use the same python where MMS installed as default.

### DIFF
--- a/mms/model_server.py
+++ b/mms/model_server.py
@@ -4,6 +4,7 @@ File to define the entry point to Model Server
 
 import os
 import subprocess
+import sys
 import tempfile
 from builtins import str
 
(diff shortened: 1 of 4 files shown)


mms/model_server.py
```diff
@@ -81,6 +82,10 @@ def start():
         cmd.append("-jar")
         cmd.append("{}/mms/frontend/model-server.jar".format(mms_home))
 
+        # model-server.jar command line parameters
+        cmd.append("-e")
```
vdantu: Why `-e` ?

frankfliu: I changed to --python

PR #629, opened 2018-10-19
Add --pre in pip install before 1.0 official release

frankfliu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ pip install mxnet-cu92mkl
 **Step 3:** Install or Upgrade MMS as follows:
 
 ```bash
-pip install -U mxnet-model-server
+pip install -U mxnet-model-server --pre
 ```
 
 **Notes:**
--- a/docs/install.md
+++ b/docs/install.md
@@ -48,13 +48,13 @@ pip install mxnet-cu90-mkl
 To install MMS for the first time, install Python, then run the following command:
 
 ```bash
-pip install mxnet-model-server
+pip install mxnet-model-server --pre
 ```
 
 To upgrade from a previous version of MMS, run:
 
 ```bash
-pip install -U mxnet-model-server
+pip install -U mxnet-model-server --pre
 ```
 
 ## Install MMS from Source Code


README.md
```diff
@@ -92,7 +92,7 @@ pip install mxnet-cu92mkl
 **Step 3:** Install or Upgrade MMS as follows:
 
 ```bash
-pip install -U mxnet-model-server
```
vdantu: I don't think we would need this. We can add a section to get the latest branch with "--pre". This could create a churn just before the release.

frankfliu: If anyone read this README today, they will run into troubles.

vdantu: The release is very close by, do you see that this churn of docs rework is required?

aaronmarkham: There's not much churn here. The docs should always be functional and accurate. Let's add --pre for the time being and point people to the 0.4 docs if they want what's in the current pip package.

README.md
```diff
@@ -92,7 +92,7 @@ pip install mxnet-cu92mkl
 **Step 3:** Install or Upgrade MMS as follows:
```
aaronmarkham:
```suggestion
**Step 3:** Install or Upgrade MMS
Option A - Install the latest nightly build of the v1.0.0 pre-release with the following:
```

README.md
```diff
@@ -92,7 +92,7 @@ pip install mxnet-cu92mkl
...
 
 ```bash
-pip install -U mxnet-model-server
+pip install -U mxnet-model-server --pre
```
aaronmarkham:
```suggestion
pip install -U mxnet-model-server --pre
```
Option B - Install v0.4.0 with the following:
```
pip install -U mxnet-model-server
```

README.md
ddavydenko: I concur that having --pre is better than not having it until public release. Does this require changes in a lot of places (that is what I can see as a churn). If it is not in a lot of places - I suggest to add --pre. We will remove it once we go public.

vdantu: If we go with a full scale change, we will need to revisit all the example docs and docker files. If we decide that this is the only file/commit, then it might be as simple as reverting a single commit.

aaronmarkham: Reverts are messy.
Let's make the change, then the day of launch, we can add another commit that fixes it. We'll probably have minor updates to make at that point anyway.

PR #631, opened 2018-10-19
Use --release parameter to build release python package.

frankfliu:
*Issue #, if available:*

*Description of changes:*


setup.py
```diff
@@ -53,8 +55,11 @@ def pypi_description():
 
 def detect_model_server_version():
     sys.path.append(os.path.abspath("mms"))
-    import mms
-    return mms.__version__
+    if "--release" in sys.argv:
+        sys.argv.remove("--release")
+        return mms.__version__.strip() + '.' + str(date.today()).replace('-', '')
+
+    return mms.__version__.strip() + '.' + str(date.today()).replace('-', '') + 'dev1'
```
vdantu: I don't know if this change is required. Because we do a manual push for released versions and its not automated. 
And, if we decide to go ahead with this change, does this work for pre release? I don't see a 'b' or a 'a' tag in the version. How would pip understand that these are not releases and just pre releases?

model-archiver/setup.py
```diff
@@ -43,20 +43,23 @@ def pypi_description():
         return df.read()
 
 
-def detect_packaging_tool_version():
+def detect_model_archiver_version():
+    if "--release" in sys.argv:
+        sys.argv.remove("--release")
+        # pylint: disable = relative-import
+        return model_archiver.__version__.strip() + '.' + str(date.today()).replace('-', '')
```
vdantu: We don't need this for model-archiver. It might just be enough to remove 'b' tag from the setup.py.

setup.py
frankfliu: With this change, we can setup the build project in codebuild. We only need click release button on codebuild.

Technically, nightly build is dev release:
https://www.python.org/dev/peps/pep-0440/#developmental-releases

--pre
Include pre-release and development versions. By default, pip only finds stable versions.

model-archiver/setup.py
frankfliu: With this change, we can also create a model-archiver-release project on codebuild.

setup.py
vdantu: It might be better to test this out with test.pypi ...

model-archiver/setup.py
vdantu: Ok.

setup.py
vdantu: I would recommend keeping it 'b' 

I see mxnet using same naming convention.

https://pypi.org/project/mxnet/#history
https://pypi.org/project/mxnet-mkl/#history

etc

frankfliu: I changed back to 'b'

Actually tensorflow is using dev: https://pypi.org/project/tf-nightly/#history

PR #639, opened 2018-10-23
Redo docker files and create nightly docker files

vdantu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docker/Dockerfile.cpu
+++ b/docker/Dockerfile.cpu
@@ -16,17 +16,17 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server \
-    && pip install mxnet-mkl
+RUN pip install --no-cache-dir mxnet-model-server \
+    && pip install --no-cache-dir mxnet-mkl==1.3.0
 
 RUN useradd -m model-server \
-    && mkdir -p /home/model-server/tmp \
-    && chown -R model-server /home/model-server
+    && mkdir -p /home/model-server/tmp
 
 COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
 COPY config.properties /home/model-server
 
-RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
+RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh \
+    && chown -R model-server /home/model-server
 
 USER model-server
 WORKDIR /home/model-server
--- a/docker/Dockerfile.nightly-cpu
+++ b/docker/Dockerfile.nightly-cpu
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:9.2-cudnn7-runtime-ubuntu18.04
+FROM ubuntu:18.04
 
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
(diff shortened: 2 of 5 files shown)


docker/Dockerfile.nightly-cpu
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    build-essential \
+    libatlas-base-dev \
+    libopencv-dev \
+    graphviz \
+    python-dev \
+    python3-dev \
```
frankfliu: We should only install python-dev since python3 doesn't work anyway.

docker/Dockerfile.cpu
```diff
@@ -16,17 +16,17 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server \
-    && pip install mxnet-mkl
+RUN pip install --no-cache-dir mxnet-model-server \
+    && pip install --no-cache-dir mxnet-mkl==1.3.0
```
piyushghai: Should we hardcode this version ? 
Would pip install mxnet-mkl not fetch whichever the latest mxnet version is ?

docker/Dockerfile.gpu
```diff
@@ -16,17 +16,17 @@ RUN apt-get update && \
     && curl -O https://bootstrap.pypa.io/get-pip.py \
     && python get-pip.py
 
-RUN pip install mxnet-model-server \
-    && pip install mxnet-cu90mkl
+RUN pip install --no-cache-dir mxnet-model-server \
+    && pip install --no-cache-dir mxnet-cu92mkl==1.3.0.post0
```
piyushghai: Same for MXNet version here.

docker/Dockerfile.cpu
piyushghai: Here's another reference that we can have a look at : 
How mxnet version is being handled for generating Docker images. 
https://github.com/apache/incubator-mxnet/pull/12791/files#diff-36826cee27ee722a559d139fc7230c28R22

docker/Dockerfile.nightly-cpu
vdantu: This change would require us to modify all our container files... Adding python3 and the required mxnet/mms in python3 (pip3 install) seems to add 300MB in CPU somewhat similar numbers in GPU. I could modify that, but I would like to take that as a separate task..

docker/Dockerfile.cpu
vdantu: This is a release Docker file, I keep it as latest in our nightly builds. I think we should hardcode the version of MXNet so that the container environment for our release-build is known.

docker/Dockerfile.gpu
vdantu: Same as above ... Once the MMS is released, I would add that version to this file as well. People who want the latest can work with the nightly builds, but for release versions, we should keep the container environment as constant as possible.

docker/Dockerfile.cpu
```diff
@@ -16,17 +16,17 @@ RUN apt-get update && \
...
-    && mkdir -p /home/model-server/tmp \
-    && chown -R model-server /home/model-server
+    && mkdir -p /home/model-server/tmp
 
 COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
 COPY config.properties /home/model-server
 
-RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
+RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh \
+    && chown -R model-server /home/model-server
```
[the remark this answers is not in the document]
frankfliu: It's better keep config file readonly from non-privileged user. Whoever want to debug it should mount volume and use their own config.properties file.

docker/Dockerfile.gpu
frankfliu: If we pass in version, we can use single docker file. Which will be easier to maintain. And we have a docker-release.sh script which hard code all the versions.

docker/Dockerfile.cpu
vdantu: @piyushghai : That link is interesting :) . Thanks for sharing it..

PR #656, opened 2018-10-31
minor fix

vdantu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/PyPiDescription.rst
+++ b/PyPiDescription.rst
@@ -52,7 +52,7 @@ or for GPU instance:
 
 ::
 
-    pip install mxnet-cu90-mkl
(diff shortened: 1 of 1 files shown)


PyPiDescription.rst
```diff
@@ -52,7 +52,7 @@ or for GPU instance:
 
 ::
 
-    pip install mxnet-cu90-mkl
+    pip install mxnet-cu90mkl
```
frankfliu: Should we change to cu92mkl ?

vdantu: Ya, I think we can. No real reason for it not to be cu92mkl.

PR #660, opened 2018-11-06
Update readme.md to 1.0 version

ThomasDelteil:


examples/gluon_character_cnn/README.md
```diff
@@ -58,44 +58,20 @@ class CharacterCNNService(GluonBaseService):
...
         self.net.hybridize()
 
-        # define _preprocess, _inference and _postprocess methods
+        # define preprocess, inference and postprocess methods
 ```
 
 As shown, the Gluon model derives from the basic gluon hybrid block. Gluon hybrid blocks, provide performance of a symbolic model with a imperative model. More on Gluon, hybrid blocks [here](https://gluon.mxnet.io/chapter07_distributed-learning/hybridize.html).
 The fully defined service file can be found under [gluon_crepe.py](gluon_crepe.py), we define `preprocess`, `inference`, `postprocess` methods in this file.
 
-## Step 3 - Check signature file
```
vdantu: Why don't we need this section?

ThomasDelteil: we don't actually use the signature file at all in this example, right now it is still loaded in the base gluon service but ideally we shouldn't even do that.

PR #677, opened 2018-11-13
Always use utf-8 to decode bytes.

frankfliu:
Tested with following command:

curl -k -X POST https://127.0.0.1:8443/predictions/noop -H "Content-Type: application/json" -d '[{"input_sentence": "好 the exchange floor as soon as ual stopped trading we <unk> for a panic said one top floor trader"}]'

This will fail before the fix and now is working fine.

### DIFF
--- a/mms/protocol/otf_message_handler.py
+++ b/mms/protocol/otf_message_handler.py
@@ -107,7 +107,7 @@ def create_load_model_response(code, message):
     msg = bytearray()
     msg += struct.pack('!i', code)
 
-    buf = message.encode()
+    buf = message.encode("utf-8")
     msg += struct.pack('!i', len(buf))
     msg += buf
     msg += struct.pack('!i', -1)  # no predictions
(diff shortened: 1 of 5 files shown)


vdantu:
The changes look good, but we might need to think of how to test it as well. Could you add a note or tests on how all these changes were tested.

mms/protocol/otf_message_handler.py
```diff
@@ -254,19 +254,19 @@ def _retrieve_input_data(conn):
...
-    content_type = _retrieve_buffer(conn, length).decode()
+    content_type = _retrieve_buffer(conn, length).decode("utf-8")
     model_input["contentType"] = content_type
 
     length = _retrieve_int(conn)
     value = _retrieve_buffer(conn, length)
 
     if content_type == "application/json":
-        model_input["value"] = json.loads(value.decode())
+        model_input["value"] = json.loads(value.decode("utf-8"))
```
vdantu: Should we change the encode() in  "create_load_model_response" to encode("utf-8") as well?

PR #679, opened 2018-11-13
QueryLog format issue fix

vrakesh:
*Issue #, if available:*
*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
@@ -26,15 +26,19 @@ public class QLogLayout extends PatternLayout {
             return null;
         }
         if (eventMessage instanceof Metric) {
+            String marketPlace = System.getenv("REALM");
             StringBuilder stringBuilder = new StringBuilder();
             Metric metric = (Metric) eventMessage;
             stringBuilder.append("HostName=").append(metric.getHostName());
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            if (marketPlace != null) {
+                stringBuilder.append("\nMarketplace=").append(marketPlace);
+            }
+            stringBuilder.append("\nStartTime=").append(metric.getTimestamp());
             stringBuilder
-                    .append("\nMetrics=")
+                    .append("\nProgram=MXNetModelServer\nMetrics=")
                     .append(metric.getMetricName())
                     .append('=')
                     .append(metric.getValue())
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/metrics/Metric.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/metrics/Metric.java
@@ -23,7 +23,7 @@ public class Metric {
 
     private static final Pattern PATTERN =
             Pattern.compile(
-                    "\\s*(\\w+)\\.(\\w+):([0-9\\-,.]+)\\|#([^|]*)\\|#hostname:([^,]+)(,(.+))?");
+                    "\\s*(\\w+)\\.(\\w+):([0-9\\-,.]+)\\|#([^|]*)\\|#hostname:([^,]+)(,(.+))(,(.+))?");
 
     @SerializedName("MetricName")
     private String metricName;
@@ -123,7 +123,8 @@ public class Metric {
         metric.setUnit(matcher.group(2));
         metric.setValue(matcher.group(3));
         metric.setHostName(matcher.group(5));
-        metric.setRequestId(matcher.group(7));
+        metric.setTimestamp(matcher.group(7));
+        metric.setRequestId(matcher.group(8));
         String dimensions = matcher.group(4);
         if (dimensions != null) {
             String[] dimension = dimensions.split(",");
@@ -157,6 +158,7 @@ public class Metric {
         if (requestId != null) {
             sb.append(",requestID:").append(requestId);
         }
+        sb.append(",timestamp:").append(timestamp);
         return sb.toString();
     }
 }
--- a/frontend/server/src/test/java/com/amazonaws/ml/mms/util/ConfigManagerTest.java
+++ b/frontend/server/src/test/java/com/amazonaws/ml/mms/util/ConfigManagerTest.java
@@ -38,7 +38,7 @@ public class ConfigManagerTest {
         metric.setMetricName(metricName);
         metric.setRequestId(requestId);
         metric.setUnit("Milliseconds");
-        metric.setTimestamp("2014-05-23 03:40");
+        metric.setTimestamp("1542157988");
         Dimension dimension = new Dimension();
         dimension.setName("Level");
         dimension.setValue("Model");
--- a/mms/metrics/metric.py
+++ b/mms/metrics/metric.py
@@ -11,7 +11,7 @@
 """
 Metric class for model server
 """
-import datetime
+import time
 import socket
 from collections import OrderedDict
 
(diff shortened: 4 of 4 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -32,9 +32,10 @@ public String format(LoggingEvent event) {
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            stringBuilder.append("\nStartTime=").append(metric.getTimestamp());
```
vdantu: I think the required fields are "StartTime", "Time", "ProgramName" and "MarketPlace". We should probably replace "EndTime" with "Time" where "Time = StartTime + Time" and MarketPlace, we will need to see how to provide it.

ddavydenko: Yes, I agree with Vamshi. Also, please note that unit for StartTime is *second* while unit for Time is *millisecond*. I think Time could be just EndTime (as currently available) - (minus) StartTime as currently available.

As for Marketplace, that is very interesting question and I don't think we can solve it quickly as it will require to introduce some customary attributes into logs that at passed all the way from configs. We need to file a backlog issue for it and not address here.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,27 +14,36 @@
 
 import com.amazonaws.ml.mms.metrics.Dimension;
 import com.amazonaws.ml.mms.metrics.Metric;
+import org.apache.log4j.Logger;
 import org.apache.log4j.PatternLayout;
 import org.apache.log4j.spi.LoggingEvent;
 
 public class QLogLayout extends PatternLayout {
+    private static final Logger logger = Logger.getLogger(QLogLayout.class);
```
frankfliu: This doesn't seems a good idea. We should avoid logging in the log call.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,27 +14,36 @@
...
     @Override
     public String format(LoggingEvent event) {
         Object eventMessage = event.getMessage();
         if (eventMessage == null) {
             return null;
         }
+        String marketPlace = "NONE";
+        try {
+            marketPlace = System.getenv("REALM");
+        } catch (Exception e) {
```
frankfliu: There should not be a expection

frankfliu: String marketPlavce = System.getEnv("REALM");
if (marketPlace == null) {
    marketPlace = "NONE";
}

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,27 +14,36 @@
...
+        }
         if (eventMessage instanceof Metric) {
             StringBuilder stringBuilder = new StringBuilder();
             Metric metric = (Metric) eventMessage;
             stringBuilder.append("HostName=").append(metric.getHostName());
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            stringBuilder.append("\nMarketPlace=").append(marketPlace);
```
frankfliu: Is MarketPlace required?
if not, we should conditionally append this section.

vrakesh: I agree MARKETPLACE is set to NONE regardless

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,27 +14,36 @@
...
     public String format(LoggingEvent event) {
         Object eventMessage = event.getMessage();
         if (eventMessage == null) {
             return null;
         }
+        String marketPlace = "NONE";
+        try {
+            marketPlace = System.getenv("REALM");
+        } catch (Exception e) {
+            logger.info("REALM env variable not set or could not be accessed");
```
vdantu: We don't need this. We should have either a value or default value for each of the metrics and just use it. We shouldn't log anything here.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,27 +14,36 @@
...
             stringBuilder.append("HostName=").append(metric.getHostName());
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            stringBuilder.append("\nMarketPlace=").append(marketPlace);
+            stringBuilder.append("\nStartTime=").append(metric.getTimestamp());
             stringBuilder
-                    .append("\nMetrics=")
+                    .append("\nProgram=MXNetModelServer\nMetrics=")
```
vdantu: Should this be called the package name? I am not sure what the expectation here is.. Usually package names are used for Program names.

vrakesh: Sure I can do a renaming for that

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -14,6 +14,7 @@
 
 import com.amazonaws.ml.mms.metrics.Dimension;
 import com.amazonaws.ml.mms.metrics.Metric;
+import org.apache.log4j.Logger;
```
frankfliu: This import is unused.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -26,15 +27,19 @@ public String format(LoggingEvent event) {
...
         if (eventMessage instanceof Metric) {
+            String marketPlace = System.getenv("REALM");
             StringBuilder stringBuilder = new StringBuilder();
             Metric metric = (Metric) eventMessage;
             stringBuilder.append("HostName=").append(metric.getHostName());
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            if (marketPlace!=null) {
```
frankfliu: format

frontend/server/src/main/java/com/amazonaws/ml/mms/util/logging/QLogLayout.java
```diff
@@ -26,15 +26,19 @@ public String format(LoggingEvent event) {
...
+            String marketPlace = System.getenv("REALM");
             StringBuilder stringBuilder = new StringBuilder();
             Metric metric = (Metric) eventMessage;
             stringBuilder.append("HostName=").append(metric.getHostName());
             if (metric.getRequestId() != null) {
                 stringBuilder.append("\nRequestId=").append(metric.getRequestId());
             }
-            stringBuilder.append("\nTimeStamp=").append(metric.getTimestamp());
+            if (marketPlace != null) {
+                stringBuilder.append("\nMarketPlace=").append(marketPlace);
```
vdantu: We don't seem to be populating the Marketplace if "REALM" is not given, shouldn't we have it set to "NONE" as a value? Doesn't metrics expect a string in return? Also, is the spelling of Marketplace correct. The reference says it should be "Marketplace" (`p` is small not capital)

vrakesh: By default if not provided it is set to NONE automatically and you are right it is Metricname

vrakesh:
Sample output in qlog format

HostName=8c85905fd915.ant.amazon.com
Marketplace=usa
StartTime=1542162515
Program=MXNetModelServer
Metrics=MemoryUtilization=68.3 Percent Level|Host 
EOE

PR #686, opened 2018-11-19
Redo batching

vdantu:
*Issue #, if available:*

Redo batching. 

*Description of changes:*

Saw the following results for LSTM with batchsize of 5 and maximum delay of 100ms

## Without changes
| Concurrency | TPS | P50 | P90 | P99 |
| ---------        | ---  | ---   | ---   | ---   |
| 5 | 49.04 | 105 | 106 | 110 |
| 8  | 79.41 | 105 | 107 | 110 |
| 10 | 105.44 | 105 | 107 | 108 |
| 25 | 1808.73 | 10 | 20 | 106 |
| 50 | 6072.04 | 7 | 9 | 51 |

## With changes
| Concurrency | TPS | P50 | P90 | P99 | P100 |
| ---------        | ---  | ---   | ---   | ---   | ---- |
| 5 | 810.55 | 6 | 6 | 8 | 37 |
| 8  | 840.93 | 11 | 12 | 15 | 26 |
| 10 | 1649.30 | 6 | 6 | 9 | 12 |
| 25 | 4225.86 | 6 | 6 | 9 | 16 |
| 50 | 6163.47 | 8 | 10 | 31 | 39 |

* I have seen variation of +- 5% in each run . I ran `ab` test 3 times and took the middle value result. 
* Varying concurrency levels (5, 8, 10, 25, 50) with 5000 messages for the above two experiments. 100 concurrent requests also showed similar results as 50 so did not test further.

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
@@ -36,49 +36,33 @@ public class BatchAggregator {
         jobs = new LinkedHashMap<>();
     }
 
-    public BaseModelRequest getRequest(String threadName) throws InterruptedException {
+    public BaseModelRequest getRequest(String threadName, WorkerState state)
+            throws InterruptedException {
         jobs.clear();
 
-        // first job is a blocking call;
-        Job job = model.nextJob(threadName);
-        if (job.isControlCmd()) {
-            RequestInput input = job.getPayload();
-            int gpuId = -1;
-            String gpu = input.getStringParameter("gpu");
-            if (gpu != null) {
-                gpuId = Integer.parseInt(gpu);
-            }
-            return new ModelLoadModelRequest(model, gpuId);
-        }
-
-        jobs.put(job.getJobId(), job);
-
-        logger.trace("get first job: {}", job.getJobId());
-
-        long maxBatchDelay = model.getMaxBatchDelay();
-        int size = model.getBatchSize() - 1;
-        long begin = System.currentTimeMillis();
-        for (int i = 0; i < size; ++i) {
-            // Only data should be consumed here.
-            job = model.nextJob(Model.DEFAULT_DATA_QUEUE, maxBatchDelay);
-            if (job == null) {
-                break;
-            }
-            jobs.put(job.getJobId(), job);
-            long end = System.currentTimeMillis();
-            maxBatchDelay -= end - begin;
-            begin = end;
-            if (maxBatchDelay <= 0) {
-                break;
-            }
-        }
-
-        logger.trace("sending jobs, size: {}", jobs.size());
-
         ModelInferenceRequest req = new ModelInferenceRequest(model.getModelName());
+
+        model.pollBatch(
+                threadName, (state == WorkerState.WORKER_MODEL_LOADED) ? 0 : Long.MAX_VALUE, jobs);
+
         for (Job j : jobs.values()) {
-            j.setScheduled();
-            req.addRequest(j.getPayload());
+            if (j.isControlCmd()) {
+                if (jobs.size() > 1) {
+                    throw new IllegalStateException(
+                            "Received more than 1 control command. "
+                                    + "Control messages should be processed/retrieved one at a time.");
+                }
+                RequestInput input = j.getPayload();
+                int gpuId = -1;
+                String gpu = input.getStringParameter("gpu");
+                if (gpu != null) {
+                    gpuId = Integer.parseInt(gpu);
+                }
+                return new ModelLoadModelRequest(model, gpuId);
+            } else {
+                j.setScheduled();
+                req.addRequest(j.getPayload());
+            }
         }
         return req;
     }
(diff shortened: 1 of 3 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -116,19 +122,49 @@ public void addFirst(Job job) {
         jobsDb.get(DEFAULT_DATA_QUEUE).addFirst(job);
     }
 
-    public Job nextJob(String threadId) throws InterruptedException {
-        return nextJob(threadId, Long.MAX_VALUE);
-    }
+    public void pollBatch(String threadId, long waitTime, Map<String, Job> jobsRepo)
+            throws InterruptedException {
+        if (jobsRepo == null || threadId == null) {
```
piyushghai: What if threadId is a blank string : "" ( a string of length 0) ? 
Can we use String.isEmpty() instead ?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -116,19 +122,49 @@ public void addFirst(Job job) {
...
+            logger.error("Invalid input given provided");
+            throw new IllegalArgumentException();
```
piyushghai: You can also pass the appropriate message in the exception perhaps ?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -116,19 +122,49 @@ public void addFirst(Job job) {
...
+        if (jobsQueue != null && !jobsQueue.isEmpty()) {
+            Job j = jobsQueue.poll(waitTime, TimeUnit.MILLISECONDS);
             if (j != null) {
-                return j;
+                jobsRepo.put(j.getJobId(), j);
+                return;
+            }
+        }
+
+        synchronized (this) {
```
piyushghai: Can you explain why you're taking lock on the entire Model object itself ?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
```diff
@@ -116,19 +122,49 @@ public void addFirst(Job job) {
...
+            long maxDelay = maxBatchDelay;
+            long begin = System.currentTimeMillis();
+            jobsQueue = jobsDb.get(DEFAULT_DATA_QUEUE);
+
+            Job j = jobsQueue.poll(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+            logger.trace("get first job: {}", Objects.requireNonNull(j).getJobId());
+
+            jobsRepo.put(j.getJobId(), j);
+            for (int i = 0; i < batchSize - 1; ++i) {
```
piyushghai: Any reason for  ```i < batchSize - 1``` and not ```i < batchSize``` ?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
frankfliu: Piyush is right, Add the message to exception and remove logging.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/BatchAggregator.java
```diff
@@ -36,50 +36,29 @@ public BatchAggregator(Model model) {
...
-            if (job == null) {
-                break;
-            }
-            jobs.put(job.getJobId(), job);
-            long end = System.currentTimeMillis();
-            maxBatchDelay -= end - begin;
-            begin = end;
-            if (maxBatchDelay <= 0) {
-                break;
+        for (Map.Entry<String, Job> j : jobs.entrySet()) {
```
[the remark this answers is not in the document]
vdantu: True, made the changes.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: Sure. We could add it. 
Its technically not needed as the threadId can't be empty, and if it is, there wouldn't be a queue for it :) ..

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: True. The logging shouldn't be there. Good catch.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: There are two kind of queues,
1. Control queue, which is per threadId. This doens't need any synchronization as each thread has its own queue and they can pull from it without any locks.
2. Data queue: This is global and shared across all threads under a model for MMS. This requries global sync because at any given time only one thread should be accessing the batch of elements.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: Good catch. This seems wrong. It doesn't affect a running system, but this reduces the batch size that can be handled by the model by 1.

vdantu:
@lufenamazon , @piyushghai . I addressed the comments.

vdantu:
@frankfliu

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/Model.java
vdantu: This should be i < batchSize -1, because the first job is inserted into the repo before we enter the job loop. This is because we have a fully-blocking wait until we receive the first message.

PR #689, opened 2018-11-20
Log load time metrics

vdantu:
*Issue #, if available:*
There is a requirement to push the load-time metrics for each model. 

## Output
After changes, the mms_metrics.log contains the following
```
HostName=hostNAme
StartTime=1542747921
Program=MXNetModelServer
Metrics=W-9000-squeezenet=2604 ms Level|Worker
EOE
HostName=hostName
StartTime=1542747921
Program=MXNetModelServer
Metrics=W-9001-squeezenet=2663 ms Level|Worker
EOE
HostName=hostName
StartTime=1542747921
Program=MXNetModelServer
Metrics=W-9003-squeezenet=2689 ms Level|Worker
EOE
```
*Description of changes:*

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
@@ -12,6 +12,8 @@
  */
 package com.amazonaws.ml.mms.wlm;
 
+import com.amazonaws.ml.mms.metrics.Dimension;
+import com.amazonaws.ml.mms.metrics.Metric;
 import com.amazonaws.ml.mms.util.ConfigManager;
 import com.amazonaws.ml.mms.util.NettyUtils;
 import com.amazonaws.ml.mms.util.codec.ModelRequestEncoder;
@@ -43,6 +45,10 @@ import org.slf4j.LoggerFactory;
 public class WorkerThread implements Runnable {
 
     static final Logger logger = LoggerFactory.getLogger(WorkerThread.class);
+    private static final org.apache.log4j.Logger loggerMmsMetrics =
+            org.apache.log4j.Logger.getLogger(ConfigManager.MMS_METRICS_LOGGER);
+
+    private Metric workerLoadTime;
 
     private static final int[] BACK_OFF = {
         0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597
@@ -97,6 +103,17 @@ public class WorkerThread implements Runnable {
         startTime = System.currentTimeMillis();
         lifeCycle = new WorkerLifeCycle(configManager, model);
         replies = new ArrayBlockingQueue<>(1);
+        workerLoadTime =
+                new Metric(
+                        "W-"
+                                + this.port
+                                + '-'
+                                + model.getModelName()
+                                        .substring(0, Math.min(model.getModelName().length(), 25)),
+                        String.valueOf(System.currentTimeMillis()),
+                        "ms",
+                        ConfigManager.getInstance().getHostName(),
+                        new Dimension("Level", "Host"));
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -97,6 +103,17 @@ public WorkerThread(
...
         lifeCycle = new WorkerLifeCycle(configManager, model);
         replies = new ArrayBlockingQueue<>(1);
+        workerLoadTime =
+                new Metric(
+                        "W-"
+                                + this.port
+                                + '-'
+                                + model.getModelName()
+                                        .substring(0, Math.min(model.getModelName().length(), 25)),
+                        Long.toString(System.currentTimeMillis()),
```
frankfliu: Stri
```suggestion
                        String.valueOf(System.currentTimeMillis()),
```

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -288,9 +305,16 @@ public void shutdown() {
 
     void setState(WorkerState newState) {
         listener.notifyChangeState(model.getModelName(), newState);
+        long timeTaken = System.currentTimeMillis() - this.startTime;
```
frankfliu:
```suggestion
        long timeTaken = System.currentTimeMillis() - startTime;
```

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -288,9 +305,16 @@ public void shutdown() {
...
         if (state != WorkerState.WORKER_TERMINATED) {
             this.state = newState;
         }
+
+        if (state == WorkerState.WORKER_MODEL_LOADED) {
+            workerLoadTime.setValue(Long.toString(timeTaken));
```
frankfliu:
```suggestion
            workerLoadTime.setValue(String.valueOf(timeTaken));
```

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -288,9 +305,16 @@ public void shutdown() {
...
+            workerLoadTime.setTimestamp(Long.toString(System.currentTimeMillis() / 1000));
```
frankfliu:
```suggestion
            workerLoadTime.setTimestamp(String.valueOf(System.currentTimeMillis() / 1000));
```

piyushghai: If the objective of the ```time/1000``` operation here is to convert to seconds, then you can use something like this : long seconds = TimeUnit.MILLISECONDS.toSeconds(millis); 

It does the same thing, but makes it easier to read the code.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
vdantu: I thought it was better to use the objects-type's toString method. Why is this better?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
vdantu: Sure. I agree. I can make this change.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
frankfliu: String.valueOf() has all the overload for most of common types.
Use String.valueOf() will be uniformed for all data types.

PR #699, opened 2018-12-05
Handle content-type returned by the backend code

vdantu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/mms/context.py
+++ b/mms/context.py
@@ -53,6 +53,15 @@ class Context(object):
     def metrics(self, metrics):
         self._metrics = metrics
 
+    def set_response_content_type(self, request_id, value):
+        self._request_processor.add_response_property(request_id, {'content-type': value})
+
+    def get_response_content_type(self, request_id):
+        response_headers = self._request_processor.get_response_header().get(request_id)
+        if response_headers is not None:
+            return response_headers.get('content-type')
+        return None
+
     def __eq__(self, other):
         return isinstance(other, Context) and self.__dict__ == other.__dict__
 
@@ -77,3 +86,6 @@ class RequestProcessor(object):
 
     def add_response_property(self, key, value):
         self._response_header[key] = value
+
+    def get_response_header(self):
+        return self._response_header
--- a/mms/protocol/otf_message_handler.py
+++ b/mms/protocol/otf_message_handler.py
@@ -45,7 +45,7 @@ def retrieve_msg(conn):
     return cmd, msg
 
 
-def create_predict_response(ret, req_id_map, message, code):
+def create_predict_response(ret, req_id_map, message, code, context=None):
     """
     Create inference response.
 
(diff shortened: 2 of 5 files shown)


mms/context.py
```diff
@@ -77,3 +80,6 @@ def report_status(self, code, reason_phrase=None):
 
     def add_response_property(self, key, value):
         self._response_header[key] = value
+
+    def get_response_property(self):
```
piyushghai: Should this not be ```get_response_headers```  ?

mms/tests/unit_tests/test_worker_service.py
```diff
@@ -40,7 +40,7 @@ def test_with_nil_request(self, service):
             service.retrieve_data_for_inference(None)
 
     def test_valid_req(self, service):
-        input_batch, req_to_id_map = service.retrieve_data_for_inference(self.data)
+        headers, input_batch, req_to_id_map = service.retrieve_data_for_inference(self.data)
```
piyushghai: Just curious, is this the only unit test that is affected by this change ?

piyushghai:
@vdantu Can you also rebase this PR with the latest master ?

mms/context.py
vdantu: The field being returned is called "response_property" and other APIs are named as such. So I wouldn't want to change the convention.

mms/tests/unit_tests/test_worker_service.py
vdantu: Yes, this is seems to be the only one affected. I think I will update this test though, to test for headers, even though this is an optional field.

mms/protocol/otf_message_handler.py
```diff
@@ -67,8 +67,15 @@ def create_predict_response(ret, req_id_map, message, code):
         msg += struct.pack("!i", len(buf))
         msg += buf
 
-        # TODO: retrieve content_type from context
-        msg += struct.pack('!i', 0)  # content_type
+        if context is None:
+            msg += struct.pack('!i', 0)  # content_type
+        else:
+            response_headers = context.request_processor.get_response_property()
+            if response_headers.get('content-type') is None:
```
vdantu: Add length check for content-type to avoid encoding empty strings.

mms/context.py
```diff
@@ -77,3 +80,6 @@ def report_status(self, code, reason_phrase=None):
 
     def add_response_property(self, key, value):
```
frankfliu: response header should be per request, we need update this API to support per request header

vdantu: True, we need that.

mms/context.py
vdantu: Actually addressed this. This makes more sense as well.

mms/service.py
```diff
@@ -60,20 +60,22 @@ def retrieve_data_for_inference(batch):
...
         for batch_idx, request_batch in enumerate(batch):
             req_id = request_batch.get('requestId').decode("utf-8")
             parameters = request_batch['parameters']
+            model_in_headers = dict()
 
             model_in = dict()
             for parameter in parameters:
                 model_in.update({parameter["name"]: parameter["value"]})
-
+                model_in_headers.update({parameter["name"]: {"content-type": parameter["contentType"]}})
```
piyushghai: Small nit here : 
Should the parameter key be ```content-type``` and not ```contentType``` ?

vdantu: `contentType` is what we get from a message from frontend. I think we can leave it as is. It is easy for us to differentiate between the message received from frontend and the actual metadata that we should use in multiple parts of the code.

PR #706, opened 2018-12-10
Update API decription to reflect error codes.

frankfliu:
1. Added 500 error code all API description.
2. Added missing error code for inference APIs.
3. Updated unittest resources.

## Testing done:


frontend/server/src/main/java/com/amazonaws/ml/mms/openapi/OpenApiUtils.java
```diff
@@ -99,10 +98,9 @@ private static Path getPingPath() {
                 "status", new Schema("string", "Overall status of the Model Server."), true);
         MediaType mediaType = new MediaType(HttpHeaderValues.APPLICATION_JSON.toString(), schema);
 
-        Response resp = new Response("200", "Model server status.", mediaType);
-
         Operation operation = new Operation("ping");
-        operation.addResponse(resp);
+        operation.addResponse(new Response("200", "Model server status", mediaType));
```
vdantu: Could we use HttpResponseStatus instead of hardcoded strings for codes?

vdantu: I agree its verbose, but wouldn't that be easier and avoids typo errors? I meant all the strings that we give in this file.

frankfliu: Is 500 more readable than String.valueOf(HttpResponseStatus.INTERNAL_SERVER_ERROR.code()) ?

vdantu: 500 is more readable, but using standard constants tell what the error (in this case INTERNAL_SERVER_ERROR) is that's being sent out and avoids any error in an accidental change in the future right?

frankfliu: We have unit-test to prevent accidental change.

PR #708, opened 2018-12-15
Make default number of workers configurable.

frankfliu:
In some condition in docker container, java is not able to report number of CPU as expected.
We should provide a way for user to configure:
1. number of netty threads: default value: 2 x availableProcessors
2. number of netty client threads: default value: 2 x availableProcessors
3. number of default model workers: default value: availableProcessors

## Testing done:


frontend/server/src/main/java/com/amazonaws/ml/mms/util/ServerGroups.java
```diff
@@ -45,9 +45,9 @@ public ServerGroups(ConfigManager configManager) {
     public final void init() {
         allChannels = new DefaultChannelGroup(GlobalEventExecutor.INSTANCE);
 
-        serverGroup = NettyUtils.newEventLoopGroup(Runtime.getRuntime().availableProcessors());
+        serverGroup = NettyUtils.newEventLoopGroup(2);
```
vdantu: What is this change for?

frankfliu: This is listener event loop thread pool, 1 is enough per listener, we are 2 listener, so it should be 2.

PR #719, opened 2019-01-03
Resolve module name conflict with mms package issue.

### LINKED ISSUE #713: [bug] cannot use a custom service file called 'service.py'
```
2018-12-20 13:04:22,838 [INFO ] W-9002-pose-stdout com.amazonaws.ml.mms.wlm.WorkerLifeCycle - TypeError: __init__() missing 6 required positional arguments: 'model_name', 'model_dir', 'manifest', 'entry_point', 'gpu', and 'batch_size'
```
etc...

It is trying to use the mms built-in service.py rather than the user-provided service.py

frankfliu:

## Issue #, if available:
Fixes #713

## Description of changes:
Avoid custom service module name conflict with mms builtin python file names.

## Testing done:
Make noop service name service.py which is the same as mms/service.py
And tested without issue.


mms/model_loader.py
```diff
@@ -112,7 +112,14 @@ def load(self, model_name, model_dir, handler, gpu_id, batch_size):
         if module_name.endswith(".py"):
             module_name = module_name[:-3]
 
+        # Remove mms dir from python path to avoid module name conflict.
+        mms_path = os.path.dirname(os.path.realpath(__file__))
+        while mms_path in sys.path:
+            sys.path.remove(mms_path)
+
         module = importlib.import_module(module_name)
+        sys.path.append(mms_path)
```
vdantu: How does this handle runtime "import"? 
What happens if we import a module at inference-runtime.
```
def inference(ctx, data):
    import service as svc
    svc.process_inference(data)
```

mms/model_loader.py
```diff
@@ -112,7 +112,14 @@ def load(self, model_name, model_dir, handler, gpu_id, batch_size):
         if module_name.endswith(".py"):
             module_name = module_name[:-3]
 
+        # Remove mms dir from python path to avoid module name conflict.
+        mms_path = os.path.dirname(os.path.realpath(__file__))
+        while mms_path in sys.path:
```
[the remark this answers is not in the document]
frankfliu: mms path appeared in the list multiple times. We need remove all occurrence from the path.

mms/model_loader.py
frankfliu: You are right, we should not add it back.
All mms code import with "mms." explicitly, we don't need mms path in the python path.
I will move the code to beginning of model_server_worker.py.

PR #731, opened 2019-01-22
Configuarable response timeout

vdantu:

## Testing done:


frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -148,6 +149,15 @@ private ConfigManager(Arguments args) {
             prop.setProperty("PYTHON_EXECUTABLE", pythonExecutable);
         }
 
+        String defaultResponseTimeout = System.getenv("DEFAULT_RESPONSE_TIMEOUT");
```
frankfliu: It doesn't seems important enough to make it an environment variable.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -431,6 +441,14 @@ private int getIntProperty(String key, int def) {
         return Integer.parseInt(value);
     }
 
+    public String getDefaultResponseTimeout() {
```
frankfliu: It's better return integer.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -431,6 +441,14 @@ private int getIntProperty(String key, int def) {
...
+        return prop.getProperty(DEFAULT_RESPONSE_TIMEOUT, "2");
+    }
+
+    public void setDefaultResponseTimeout(String timeout) {
```
frankfliu: We don't need a setter.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -510,6 +528,7 @@ private static int getAvailableGpu() {
         private String pythonExecutable;
         private String modelStore;
         private String[] models;
+        private String defaultResponseTO;
```
frankfliu: better use defaultResponseTimeout

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -550,13 +570,24 @@ public static Options getOptions() {
                             .argName("MODELS-STORE")
                             .desc("Model store location where models can be loaded.")
                             .build());
+            options.addOption(
```
frankfliu: Not necessary to make it a cli option

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -167,6 +168,10 @@ private void handleRegisterModel(ChannelHandlerContext ctx, QueryStringDecoder d
         int initialWorkers = NettyUtils.getIntParameter(decoder, "initial_workers", 0);
         boolean synchronous =
                 Boolean.parseBoolean(NettyUtils.getParameter(decoder, "synchronous", null));
+        String responseTimeout = NettyUtils.getParameter(decoder, "response_timeout", null);
```
frankfliu: We need update OpenApi document if we make this api change.

frankfliu: int responseTimeout = NettyUtils.getIntParameter(decoder, ...);

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: True, but I was trying to keep it consistent with our configurations. This way, customers should be able to configure anything in the config.properties through either CLI or environment variables. This is similar to other configurations that we have.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: The way I have the code is, we consume this as String in the beginning, but convert to Integer only once per model loading. Would this cause any issues?

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: We would need the setter because this can come through CLI.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: Will change this

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: I had similar reasoning here as the Environment variable. This keeps MMS config consistent.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
vdantu: Will make this change.

vdantu: Similar to above comment, I wanted to make this INT transform only before its needed. I can make this change though.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
frankfliu: We should not provide multiple way to do the same thing.
All the options exposed to CLI or Env should have a strong justification.
Currently we provide 4 options in CLI:
mms-config-file: config.properties cannot config itself, this is must have
python: this information is dynamic, java can't detect by itself
models/model-store: this is for backward compatible.

vdantu: After discussion, I agree we don't need this as CLI/Environment variable. I will remove it for now. We can take it up when we get a requirement for it.

frontend/server/src/main/java/com/amazonaws/ml/mms/openapi/OpenApiUtils.java
```diff
@@ -321,6 +321,12 @@ private static Operation getRegisterOperation() {
                         "integer",
                         "100",
                         "Maximum delay for batch aggregation, default: 100."));
+        operation.addParameter(
+                new QueryParameter(
+                        "response_timeout",
+                        "integer",
+                        "2",
+                        "Maximum time, in minutes, the Model Server waits for a response from the model inference code, default: 2."));
```
frankfliu: I am wondering if this should be seconds.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -167,6 +168,11 @@ private void handleRegisterModel(ChannelHandlerContext ctx, QueryStringDecoder d
         int initialWorkers = NettyUtils.getIntParameter(decoder, "initial_workers", 0);
         boolean synchronous =
                 Boolean.parseBoolean(NettyUtils.getParameter(decoder, "synchronous", null));
+        int responseTimeout =
```
frankfliu: This can be simplified as: 
        int responseTimeout = Integer.parseInt(NettyUtils.getParameter(decoder, "response_timeout", ConfigManager.getInstance().getDefaultResponseTimeout());

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
```diff
@@ -550,6 +555,13 @@ public static Options getOptions() {
...
+                    Option.builder("t")
```
frankfliu: As we discussed, I don't think introduce a cli option is a must have.

frontend/server/src/main/java/com/amazonaws/ml/mms/openapi/OpenApiUtils.java
vdantu: Agreed, I was thinking the same. I will change this to seconds anyways.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
vdantu: Done

PR #733, opened 2019-01-24
Consume Enginename as a string

vdantu:

## Testing done:

### DIFF
--- a/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/LegacyManifest.java
+++ b/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/LegacyManifest.java
@@ -113,7 +113,7 @@ public class LegacyManifest {
             Object engineVersion = engine.get("MXNet");
             if (engineVersion instanceof Number) {
                 Manifest.Engine eng = new Manifest.Engine();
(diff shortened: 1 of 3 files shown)


frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/LegacyManifest.java
```diff
@@ -113,7 +113,7 @@ public Manifest migrate() throws InvalidModelException {
             Object engineVersion = engine.get("MXNet");
             if (engineVersion instanceof Number) {
                 Manifest.Engine eng = new Manifest.Engine();
-                eng.setEngineName(Manifest.EngineType.MX_NET);
+                eng.setEngineName("MXNet");
```
frankfliu: You don't need change it to String, the only the you need change is:
if (engineType.value.equals(value)) {

change it to:
if (engineType.value.equalsIgnoreCase(value)) {

vdantu: I made these changes because MMS doesn't use Engine and this gives flexibility for customers to define the name that they want to use. Also, this removes MMS from needing to make code changes when ever we add support for newer frameworks. I will drop by and discuss this in person.

PR #737, opened 2019-01-25
Update benchmarkAI script.

frankfliu:

## Testing done:

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -50,19 +50,21 @@ JMX_MULTIPLE_MODELS_LOAD_PLAN = 'multipleModelsLoadPlan.jmx'
 JMX_GRAPHS_GENERATOR_PLAN = 'graphsGenerator.jmx'
 
 # Listing out the models tested
-MODEL_RESNET_18 = 'resnet-18'
-MODEL_SQUEEZE_NET = 'squeezenet'
+MODEL_RESNET_50v1 = 'resnet-50v1'
 MODEL_LSTM_PTB = 'lstm_ptb'
 MODEL_NOOP = 'noop-v1.0'
 
-
 MODEL_MAP = {
-    MODEL_SQUEEZE_NET: (JMX_IMAGE_INPUT_MODEL_PLAN, {'url': 'https://s3.amazonaws.com/model-server/models/squeezenet_v1.1/squeezenet_v1.1.model', 'model_name': MODEL_SQUEEZE_NET, 'input_filepath': 'kitten.jpg'}),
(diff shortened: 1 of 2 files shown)


benchmarks/benchmark.py
```diff
@@ -434,40 +430,52 @@ def run_benchmark():
...
 
 
-def modify_config_props_for_mms(pargs):
-    shutil.copyfile(CONFIG_PROP_TEMPLATE, CONFIG_PROP)
-    with open(CONFIG_PROP, 'a') as f:
-        f.write('\nnumber_of_netty_threads=32')
-        f.write('\njob_queue_size=1000')
-        if pargs.gpus:
-            f.write('\nnumber_of_gpu={}'.format(pargs.gpus[0]))
+def modify_config_props_for_mms(p_args, out_dir, model_name, model_url):
```
zachgk: You should just use pargs instead of having p_args in only this one place

benchmarks/benchmark.py
```diff
@@ -187,24 +186,17 @@ def run_single_benchmark(jmx, jmeter_args=dict(), threads=100, out_dir=None):
...
         docker_path = 'awsdeeplearningteam/mxnet-model-server:nightly-mxnet-gpu' \
             if pargs.gpus else 'awsdeeplearningteam/mxnet-model-server:nightly-mxnet-cpu'
         if pargs.docker:
-            container = 'mms_benchmark_{}'.format(pargs.docker[0].split('/')[1])
             docker_path = pargs.docker[0]
-        run_process("{} rm -f {}".format(docker, container))
-        docker_run_call = "{} run --name {} -p 8080:8080 -p 8081:8081 -v {}:{} -itd {}".format(docker, container, MMS_BASE, DOCKER_MMS_BASE, docker_path)
+        run_process("docker rm -f mms")
+        run_process("docker pull {}".format(docker_path))
+        docker_run_call = "{} run --name mms -p 8080:8080 -p 8081:8081 -v {}:/mms -itd {}" \
```
zachgk: Since you are switching to docker commands instead of using the docker variable everywhere, maybe you should change to using `nvidia_docker = "--runtime=nvidia" if pargs.gpus else ""` and then add that to docker_run_call and hardcode the word docker. Also, doesn't docker run automatically pull if necessary?

benchmarks/benchmark.py
frankfliu: p_args is a parameter which conflict with pargs. The PyCharm give warning about this conflict.

benchmarks/benchmark.py
frankfliu: docker run doesn't pull the latest image if already have on in local.

benchmarks/benchmark.py
zachgk: pargs is a global, so you could remove p_args from the arguments and access the original pargs directly

PR #741, opened 2019-01-26
Update mms version.

frankfliu:
Update mms and model-archiver version.
Without this change pip --pre will not get latest preview version.

## Testing done:

### DIFF
--- a/model-archiver/setup.py
+++ b/model-archiver/setup.py
@@ -70,9 +70,10 @@ if __name__ == '__main__':
         packages=pkgs,
         install_requires=requirements,
         extras_require={
-            'mxnet-mkl': ['mxnet-mkl==1.2.0'],
-            'mxnet-cu90mkl': ['mxnet-cu90mkl==1.2.0'],
-            'mxnet': ['mxnet==1.2'],
+            'mxnet-mkl': ['mxnet-mkl==1.3.1'],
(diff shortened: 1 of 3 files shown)


model-archiver/setup.py
```diff
@@ -70,9 +70,9 @@ def detect_model_archiver_version():
         packages=pkgs,
         install_requires=requirements,
         extras_require={
-            'mxnet-mkl': ['mxnet-mkl==1.2.0'],
-            'mxnet-cu90mkl': ['mxnet-cu90mkl==1.2.0'],
-            'mxnet': ['mxnet==1.2'],
+            'mxnet-mkl': ['mxnet-mkl==1.3.1'],
+            'mxnet-cu90mkl': ['mxnet-cu90mkl==1.3.1'],
```
vdantu: Should we change this to cu92mkl? I think our containers use cu92mkl, https://github.com/awslabs/mxnet-model-server/blob/master/docker/Dockerfile.gpu#L22

frankfliu: This something that I'm not sure. I didn't verify it for model-archiver

vdantu: Model Archiver seems to use MXNet for converting ONNX models to MXNet models. I think if this works for cu90mkl it should work for cu92mkl as well, right?

PR #752, opened 2019-02-13
Added handling for accept headers

vdantu:

## Issue #, if available:
Added accept header handling to MMS.

## Testing done:

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
@@ -154,6 +154,8 @@ public class InferenceRequestHandler extends HttpRequestHandler {
         }
 
         CharSequence contentType = HttpUtil.getMimeType(req);
+        Map<String, String> acceptMap = NettyUtils.getAcceptHeader(req);
+
         if (HttpPostRequestDecoder.isMultipart(req)
                 || HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.contentEqualsIgnoreCase(
                         contentType)) {
@@ -171,6 +173,7 @@ public class InferenceRequestHandler extends HttpRequestHandler {
             }
         } else {
             byte[] content = NettyUtils.getBytes(req.content());
+            inputData.setHeaders(acceptMap);
             inputData.addParameter(new InputParameter("body", content, contentType));
         }
         return inputData;
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
@@ -154,7 +154,9 @@ public class InferenceRequestHandler extends HttpRequestHandler {
         }
 
         CharSequence contentType = HttpUtil.getMimeType(req);
-        Map<String, String> acceptMap = NettyUtils.getAcceptHeader(req);
+        for (Map.Entry<String, String> entry : req.headers().entries()) {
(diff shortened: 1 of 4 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/util/NettyUtils.java
```diff
@@ -236,6 +239,42 @@ public static int getIntParameter(QueryStringDecoder decoder, String key, int de
         }
     }
 
+    public static Map<String, String> getAcceptHeader(HttpMessage message) {
```
frankfliu: 1. I think SM only care about "Accept" header, other headers are not relevant.
2. since inputData already takes a map, why not put all headers in it?

frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
```diff
@@ -171,6 +173,7 @@ private static RequestInput parseRequest(
             }
         } else {
             byte[] content = NettyUtils.getBytes(req.content());
+            inputData.setHeaders(acceptMap);
```
frankfliu: Headers should be set even for multipart case.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
```diff
@@ -154,6 +154,8 @@ private static RequestInput parseRequest(
         }
 
         CharSequence contentType = HttpUtil.getMimeType(req);
+        Map<String, String> acceptMap = NettyUtils.getAcceptHeader(req);
```
frankfliu:
```suggestion
        Map<String, String> headers = inputData.getHeaders();
        for(Iterator<Map.Entry<String, String>> it = req.headers().iteratorAsString(); it.hasNext();) {
            Map.Entry<String, String> entry = it.next();
            headers.put(entry.getKey(), entry.getValue());
        }
```

frontend/server/src/main/java/com/amazonaws/ml/mms/util/NettyUtils.java
vdantu: I just wanted to handle all the accept headers. SM can use only Accept header for now and expand it to others as and when they need to.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
vdantu: Yes, this is the discussion I had with you. I thought we concluded that Accept headers in multipart isn't used as much. I can add it, but handling that is pretty weird. Right now the Accept headers are used at a Request level, but if we handle multipart, then Accept headers will become at a input parameter level.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/InferenceRequestHandler.java
vdantu: Sure. I can make this change.

frontend/server/src/main/java/com/amazonaws/ml/mms/util/messages/RequestInput.java
```diff
@@ -46,6 +46,10 @@ public void setHeaders(Map<String, String> headers) {
         this.headers = headers;
     }
 
+    public void updateHeaders(String key, String val) {
+        this.headers.put(key, val);
```
frankfliu: unnecessary this modifier.

vdantu: Done

PR #756, opened 2019-02-22
MMS in single model mode

vdantu:

## Testing done:

### DIFF
--- a/docker/config.properties
+++ b/docker/config.properties
@@ -19,4 +19,5 @@ management_address=http://0.0.0.0:8081
 # keystore_type=PKCS12
 # private_key_file=src/test/resources/key.pem
 # certificate_file=src/test/resources/certs.pem
+# single_model_mode=false
 # blacklist_env_vars=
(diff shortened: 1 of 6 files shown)


docker/config.properties
```diff
@@ -19,4 +19,5 @@ management_address=http://0.0.0.0:8081
 # keystore_type=PKCS12
 # private_key_file=src/test/resources/key.pem
 # certificate_file=src/test/resources/certs.pem
+# single_model_mode=false
```
frankfliu: I don't really like have a single model mode concept.
This make things complicated:
1. Hard to enforce initial model loading
2. Hard to communicate to user what this exactly mean and what's the use case.
3. Require lock to preventing multiple registration at the same time.
4. I can foresee SM will ask a mix mode in future, which will create more issues.

I prefer define a default model concept instead of single model model.
If invocation call without model name, use default model, if no default model define, fail the request.

vdantu: Sure, we could do it a default-model way as well. But I see it opening a new can of worms. For example, one of the issues with default model is that the service owner might not know what the default model is, when they are launching MMS. So we can't have this option at startup time. Now, if we add it to register model command, then there are possibilities of having unknown behavior when two models are defined as "default" models. We will have to synchronize these calls again. 

I need to handle some conditions in this code that I have missed, but this configuration tells customers that in this mode MMS handles only one model and no more. This way, we can define "invocations" as a single model API and predictions as a multi-model API. This clearly distinguishes the behavior of MMS.  For the errors, I have added messages that are thrown. I would like to understand what "mix-mode" means. We can have a discussion about this tomorrow.

vdantu:
Taking this PR down. I will resubmit a new PR.

PR #760, opened 2019-02-28
Fix bug #759, reset previous error in handle function.

### LINKED ISSUE #759: A bug occurred while processing an error request
When I send an erroneous request with python's requests package, I get an error "'NoneType' object is not iterable". But when I send a correct request, I still get this error. So I suspect that the error flag in the service program has not been reset. so I added self.error=None at the beginning of the data receiving function in mxnet_vision_server.py, and the problem disappeared.

frankfliu:
Fixes #759

## Testing done:


examples/model_service_template/model_handler.py
```diff
@@ -73,6 +73,7 @@ def handle(self, data, context):
         :param context: model server context
         :return: list of outputs to be send back to client
         """
+        self.error = None  # reset earlier errors
```
ddavydenko: I am curious, why do we need to explicitly reset error state? If this state is shared between requests then resetting it like this might not be safe. If it is not shared and each request has its own state for error - it is not clear why this explicit set to None is necessary.

frankfliu: It's shared between multiple requests.
Each worker loads a model and created Service object. error is a member variable of the service object, so it need to be reset on each request.
Each work can only handle one request at a time, so it's safe to reuse service object.

vdantu: I am wondering the use of these template files now. We have had a few folks in the MMS user community who were confused why there were these many classes and inheritances. Should we just revamp this code and have all of this in one file? I was planning on replacing all these files with a single service file.

ddavydenko: Oh, ok, I forgot that each worker can handle only one request at a time, my bad. I approve this PR then. Re: Vamshi's comment on it, I think it would be nice to revamp the code and make it a bit simpler (even if it becomes less OOP-like).

vdantu: Hmm. I re-read the comment. I think I will raise a new PR with revamping the model-files. I think we should have this fix in until then. Just curious, is changing all the model-zoo models, which use this file, also in the scope for this PR? Or should we take a separate task to do that?

vdantu:
Please make a note of the work to publish these changes to model-zoo, if its not in the scope of this PR.

PR #761, opened 2019-03-03
 Handle default invocations model_name

vdantu:

## Testing done:

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
@@ -132,7 +132,7 @@ public class ManagementRequestHandler extends HttpRequestHandler {
         resp.setMaxBatchDelay(model.getMaxBatchDelay());
         resp.setMaxWorkers(model.getMaxWorkers());
         resp.setMinWorkers(model.getMinWorkers());
-        resp.setModelRegisteredAtStartup(modelManager.getStartupModels().contains(modelName));
+        resp.setLoadedAtStartup(modelManager.getStartupModels().contains(modelName));
         Manifest manifest = model.getModelArchive().getManifest();
         Manifest.Engine engine = manifest.getEngine();
         if (engine != null) {
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
@@ -132,6 +132,8 @@ public class ManagementRequestHandler extends HttpRequestHandler {
         resp.setMaxBatchDelay(model.getMaxBatchDelay());
         resp.setMaxWorkers(model.getMaxWorkers());
         resp.setMinWorkers(model.getMinWorkers());
+        resp.setModelInitialization(
+                modelManager.getStartupModels().contains(modelName) ? "At startup" : "At runtime");
         Manifest manifest = model.getModelArchive().getManifest();
         Manifest.Engine engine = manifest.getEngine();
         if (engine != null) {
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
@@ -132,8 +132,7 @@ public class ManagementRequestHandler extends HttpRequestHandler {
         resp.setMaxBatchDelay(model.getMaxBatchDelay());
         resp.setMaxWorkers(model.getMaxWorkers());
         resp.setMinWorkers(model.getMinWorkers());
-        resp.setModelInitialization(
-                modelManager.getStartupModels().contains(modelName) ? "At startup" : "At runtime");
+        resp.setModelRegisteredAtStartup(modelManager.getStartupModels().contains(modelName));
         Manifest manifest = model.getModelArchive().getManifest();
         Manifest.Engine engine = manifest.getEngine();
         if (engine != null) {
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
@@ -25,6 +25,7 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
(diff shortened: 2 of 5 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -41,13 +42,15 @@
     private ConfigManager configManager;
     private WorkLoadManager wlm;
     private ConcurrentHashMap<String, Model> models;
+    private ConcurrentLinkedDeque<Model> startupModels;
```
frankfliu: ArrayList<Model> should be enough, this is only added during startup time.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -102,6 +114,11 @@ public ModelArchive registerModel(
             throw new BadRequestException("Model " + modelName + " is already registered.");
         }
         logger.info("Model {} loaded.", model.getModelName());
+
+        if (startupModel) {
+            ModelManager.getInstance().setStartupModels(model);
```
frankfliu:
```suggestion
            startupModels.add(model);
```

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -193,4 +210,12 @@ public boolean scaleRequestStatus(String modelName) {
     public void submitTask(Runnable runnable) {
         wlm.scheduleAsync(runnable);
     }
+
+    public ConcurrentLinkedDeque<Model> getStartupModels() {
+        return startupModels;
+    }
+
+    public void setStartupModels(Model model) {
```
frankfliu: Should not expose setter.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: True. I agree.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: Ya, true. We shouldn't expose public setter.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: I don't see a need for model. I redid this PR with just the model-name. It works for most cases and has less references of "model" object itself. This may be better from memory clean up perspective

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -132,6 +132,8 @@ private void handleDescribeModel(ChannelHandlerContext ctx, String modelName)
         resp.setMaxBatchDelay(model.getMaxBatchDelay());
         resp.setMaxWorkers(model.getMaxWorkers());
         resp.setMinWorkers(model.getMinWorkers());
+        resp.setModelInitialization(
```
frankfliu: the response is a json, the value is expected to be parsed by program.
It's better use boolean type.
The human readable string can cause localization issue in future.
And sole to key name is part of API. It's better name it as something like "StartupModel"

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -116,6 +122,7 @@ public boolean unregisterModel(String modelName) {
         model.setMaxWorkers(0);
         wlm.modelChanged(model);
         model.getModelArchive().clean();
+        startupModels.remove(modelName);
```
frankfliu: You don't have to remove it.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -193,4 +200,8 @@ public boolean scaleRequestStatus(String modelName) {
     public void submitTask(Runnable runnable) {
         wlm.scheduleAsync(runnable);
     }
+
+    public Set<String> getStartupModels() {
+        return startupModels;
```
frankfliu: If you use a list, you can access it by index.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -41,21 +42,25 @@
     private ConfigManager configManager;
     private WorkLoadManager wlm;
     private ConcurrentHashMap<String, Model> models;
+    private Set<String> startupModels;
     private ScheduledExecutorService scheduler;
 
-    private ModelManager(ConfigManager configManager, WorkLoadManager wlm) {
+    private ModelManager(
+            ConfigManager configManager, WorkLoadManager wlm, Set<String> startupModels) {
```
frankfliu: You don't need to pass startModels in constructor. You can create it in the constructor, it will be much cleaner.

In the initialization code, you simplely call: modelManager.getStartupModel().add(modelName).

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: We do. This can be deleted and it can come back through register API. We need to clean up.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: I use a set to make it easy for show commands.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: This seems cleaner as we don't need any API. This can be set up only during initialization of ModelManager and no where else.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
vdantu: Aah, true. I can use that as well. I thought I could be more descriptive, but we don't need to do it. Probably uses less space as well.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
frankfliu: You already have getStartupModels() in ModelManager.

vdantu: Oh, that way. I think we should change this getter then. We shouldn't have a mean to write to this object post startup. I will make changes.

vdantu: I will fall back to mutable sets. Trying to make them immutable to "add" seems more complicated.

PR #763, opened 2019-03-05
Configurable response buffer size

### LINKED ISSUE #762: Exception thrown on large responses
When returning a large response from the backend worker (eg. tens of images at once), I'm seeing the following exception:

```
io.netty.handler.codec.CorruptedFrameException: Message size exceed limit: 6881420
	at com.amazonaws.ml.mms.util.codec.CodecUtils.readLength(CodecUtils.java:34)
	at com.amazonaws.ml.mms.util.codec.ModelResponseDecoder.decode(ModelResponseDecoder.java:60)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:428)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:392)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:359)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInactive(ByteToMessageDecoder.java:342)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:231)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:224)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelInactive(DefaultChannelPipeline.java:1429)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:231)
	at io.netty.channel.DefaultChannelPipeline.fireChannelInactive(DefaultChannelPipeline.java:947)
	at io.netty.channel.AbstractChannel$AbstractUnsafe$8.run(AbstractChannel.java:822)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:309)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:884)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

The exception originates from [this line](https://github.com/awslabs/mxnet-model-server/blob/master/frontend/server/src/main/java/com/amazonaws/ml/mms/util/codec/ModelResponseDecoder.java#L60) in `ModelResponseDecoder`. It appears that any response larger than the max buffer size is dropped.

Our use case is a video processing model which returns several 100s of frames in a single prediction (~ 50MB when serialized).

I'll be happy to open a PR to address this with some guidance on what you think is the best approach here (should we make the buffer size configurable? implement stream-like reading?)

erandagan:
Resolves #762 

## Motivation
Enable models which return very large responses (eg. images / videos) to run within MMS.

## Description of changes
This PR adds a new optional configuration parameter, `backend_response_buffer_size`, which controls the max size allocated to a single worker response. The buffer size was previously hard-coded at 6553500 bytes, which now serves as the default value.

## Testing done
- Tested with an artificially large prediction response to verify the buffer size is used correctly.
- Ran CI locally.

### DIFF
--- a/docker/advanced_settings.md
+++ b/docker/advanced_settings.md
@@ -307,6 +307,7 @@ Notes on a couple of the parameters:
 * **keystore_type** - Store of cryptographic keys and certificates
 * **private_key_file** - Location of the private key file
 * **certificate_file** - Location of the certificate file
+* **backend_response_buffer_size** - The maximum buffer size the frontend allocates for a worker response, in bytes.
 
 in the range of 0 .. (num-gpu-1) in a round-robin fashion. **By default MMS uses all the available GPUs but this parameter can be configured if user want to use only few of them**.
 
@@ -325,4 +326,5 @@ in the range of 0 .. (num-gpu-1) in a round-robin fashion. **By default MMS uses
 # keystore_type=PKCS12
 # private_key_file=src/test/resources/key.pem
 # certificate_file=src/test/resources/certs.pem
+# backend_response_buffer_size=6553500
 ```
--- a/docker/advanced_settings.md
+++ b/docker/advanced_settings.md
@@ -326,5 +326,5 @@ in the range of 0 .. (num-gpu-1) in a round-robin fashion. **By default MMS uses
 # keystore_type=PKCS12
 # private_key_file=src/test/resources/key.pem
 # certificate_file=src/test/resources/certs.pem
-# backend_response_buffer_size=6553500
+# max_response_size=6553500
 ```
--- a/docker/advanced_settings.md
+++ b/docker/advanced_settings.md
@@ -307,7 +307,7 @@ Notes on a couple of the parameters:
 * **keystore_type** - Store of cryptographic keys and certificates
 * **private_key_file** - Location of the private key file
 * **certificate_file** - Location of the certificate file
-* **backend_response_buffer_size** - The maximum buffer size the frontend allocates for a worker response, in bytes.
+* **max_response_size** - The maximum buffer size the frontend allocates for a worker response, in bytes.
 
 in the range of 0 .. (num-gpu-1) in a round-robin fashion. **By default MMS uses all the available GPUs but this parameter can be configured if user want to use only few of them**.
 
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/util/ConfigManager.java
@@ -86,6 +86,8 @@ public final class ConfigManager {
     private static final String CERTIFICATE_FILE = "certificate_file";
     private static final String PRIVATE_KEY_FILE = "private_key_file";
 
+    private static final String BACKEND_RESPONSE_BUFFER_SIZE = "backend_response_buffer_size";
+
     private Pattern blacklistPattern;
     private Properties prop;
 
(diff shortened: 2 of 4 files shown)


frankfliu:
Please run following command to reformat your java code:
```
gradle fJ
```

docker/advanced_settings.md
```diff
@@ -307,6 +307,7 @@ Notes on a couple of the parameters:
 * **keystore_type** - Store of cryptographic keys and certificates
 * **private_key_file** - Location of the private key file
 * **certificate_file** - Location of the certificate file
+* **backend_response_buffer_size** - The maximum buffer size the frontend allocates for a worker response, in bytes.
```
frankfliu: Can we change it to max_response_size?

backend is a bit ambiguous.

vdantu:
@erandagan : Thanks for raising a PR. It is highly appreciated :) . Could you please also address the comments so that we can merge this PR.

docker/advanced_settings.md
erandagan: Should we split them to 2 attributes then (`max_backend_response_size` and `max_response_size`) or just define a single `max_response_size` which both the frontend and the backend will share?

vdantu: I also recommend not using "backend" in the configuration parameters. Its better to leave it as "max-message-size" or something generic like that. I see value in having separate configuration parameters for backend request and response, from memory point of view, but I also feel we are adding too many configuration parameters. I would recommend having only one parameter for now and later on, based on requirements, we could add more.

erandagan: Sounds reasonable, renamed to `max_response_size`.

docker/advanced_settings.md
```diff
@@ -325,4 +326,5 @@ in the range of 0 .. (num-gpu-1) in a round-robin fashion. **By default MMS uses
 # keystore_type=PKCS12
 # private_key_file=src/test/resources/key.pem
 # certificate_file=src/test/resources/certs.pem
+# backend_response_buffer_size=6553500
```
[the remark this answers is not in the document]
vdantu: Can we make this max_message_size? This is used both for sending request and receiving response.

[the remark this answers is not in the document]
erandagan: @vdantu Well, as long as we don't foresee any use case of having a different configuration for the request size and the response size, it does seem like a better name. What do you think? Should we support that use case?

Also, note we don't use that value for requests (just yet)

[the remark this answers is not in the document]
ddavydenko: I agree with @erandagan. Response size limit should not affect request size limit and vice versa. I suggest to keep this very simple: `max_request_size` and `max_response_size` with both params applicable in respective cases for backend and frontend to the same extent.

PR #766, opened 2019-03-08
Nightly base containers

vdantu:

## Testing done:

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -68,41 +68,38 @@ docker rm -f mms
 
 ## Available pre-built continers
 We have 5 containers tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
-1. *latest*: This will be the same as `mxnet-model-server:1.0.0-mxnet-cpu`. This tag will be available after an official release.
-2. *1.0.1-mxnet-cpu*: This will be the official CPU container based on the [Dockerfile.cpu](https://github.com/awslabs/mxnet-model-server/blob/master/docker/Dockerfile.cpu). This tag will be available after an official 1.0 release.
-3. *1.0.1-mxnet-gpu*: This will be the official GPU container based on the [Dockerfile.gpu](https://github.com/awslabs/mxnet-model-server/blob/master/docker/Dockerfile.gpu). This tag will be available after an official 1.0 release.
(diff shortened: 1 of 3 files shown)


docker/README.md
```diff
@@ -68,41 +68,38 @@ docker rm -f mms
 
 ## Available pre-built continers
 We have 5 containers tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
```
ddavydenko:
```suggestion
We have 8 containers tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
```

docker/README.md
```diff
@@ -68,41 +68,38 @@ docker rm -f mms
 
 ## Available pre-built continers
```
ddavydenko: This whole section talks about a lot of containers in future tense, as in 'will be built'. I believe these containers are already available so we should update language to reflect that correctly.

docker/README.md
```diff
@@ -67,7 +67,7 @@ docker rm -f mms
 ```
 
 ## Available pre-built continers
-We have 5 containers tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
+We have 8 containers tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
```
piyushghai: I see only 7 tags on Docker hub. 
Am I missing some tag ? 
https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/tags

ddavydenko: Good catch, @piyushghai ! @vdantu let's please make sure to match our document with what we currently have in dockerhub.

vdantu: I don't think we should touch this document for every release of MMS. I have a couple of lines there calling out that we have one released container for every release of MMS since 1.0.0

docker/README.md
vdantu: Updated this.

docker/README.md
ddavydenko:
```suggestion
We have following container tags available on [Docker Hub](https://hub.docker.com/r/awsdeeplearningteam/mxnet-model-server/).
```
I propose to remove number of tags available so that we won't have to update this file every release but also so that we won't have discrepancy in number we announce and actual available containers.

PR #773, opened 2019-03-15
Model archiver changes

vdantu:

## Description of changes:
Proposal to modify the model archiver tool to create archives which can be hosted on sagemaker

## Testing done:

### DIFF
--- a/model-archiver/model_archiver/arg_parser.py
+++ b/model-archiver/model_archiver/arg_parser.py
@@ -74,6 +74,15 @@ class ArgParser(object):
                                         'parameter. If --export-path is not specified, the file will be saved in the '
                                         'current working directory. ')
 
+        parser_export.add_argument('--archive-format',
+                                   required=False,
+                                   type=str,
+                                   default="mar",
+                                   choices=["tgz","mar"],
+                                   help='The format in which the model artifacts are archived\n'
+                                        'tgz: This is the format in which Sagemaker models are consumed\n'
+                                        'mar: This is the format in which models are consumed in MMS\n')
+
         parser_export.add_argument('-f', '--force',
                                    required=False,
                                    action='store_true',
--- a/model-archiver/model_archiver/arg_parser.py
+++ b/model-archiver/model_archiver/arg_parser.py
@@ -77,11 +77,11 @@ class ArgParser(object):
         parser_export.add_argument('--archive-format',
                                    required=False,
                                    type=str,
-                                   default="mar",
-                                   choices=["tgz", "mar"],
+                                   default="default",
+                                   choices=["tgz", "default"],
                                    help='The format in which the model artifacts are archived\n'
-                                        'tgz: This is the format in which Sagemaker models are consumed\n'
-                                        'mar: This is the format in which models are consumed in MMS\n')
+                                        'tgz: This creates the model-archive in <model-name>.tar.gz format\n'
+                                        'default: This creates the model-archive in <model-name>.mar format\n')
 
         parser_export.add_argument('-f', '--force',
                                    required=False,
--- a/model-archiver/model_archiver/arg_parser.py
+++ b/model-archiver/model_archiver/arg_parser.py
@@ -78,7 +78,7 @@ class ArgParser(object):
                                    required=False,
                                    type=str,
                                    default="mar",
-                                   choices=["tgz","mar"],
+                                   choices=["tgz", "mar"],
                                    help='The format in which the model artifacts are archived\n'
                                         'tgz: This is the format in which Sagemaker models are consumed\n'
                                         'mar: This is the format in which models are consumed in MMS\n')
--- a/model-archiver/model_archiver/arg_parser.py
+++ b/model-archiver/model_archiver/arg_parser.py
@@ -79,9 +79,13 @@ class ArgParser(object):
                                    type=str,
                                    default="default",
                                    choices=["tgz", "default"],
-                                   help='The format in which the model artifacts are archived\n'
-                                        'tgz: This creates the model-archive in <model-name>.tar.gz format\n'
-                                        'default: This creates the model-archive in <model-name>.mar format\n')
+                                   help='The format in which the model artifacts are archived.\n'
+                                        'tgz: This creates the model-archive in <model-name>.tar.gz format.'
+                                        'If platform hosting MMS requires model-artifacts to be in \".tar.gz\"'
+                                        ' use this option.\n'
+                                        'default: This creates the model-archive in <model-name>.mar format.'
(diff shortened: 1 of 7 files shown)


model-archiver/model_archiver/arg_parser.py
```diff
@@ -74,6 +74,15 @@ def export_model_args_parser():
                                         'parameter. If --export-path is not specified, the file will be saved in the '
                                         'current working directory. ')
 
+        parser_export.add_argument('--archive-format',
+                                   required=False,
+                                   type=str,
+                                   default="mar",
+                                   choices=["tgz", "mar"],
```
vdantu: Should we change this option to "sagemaker" and "mms" instead of "tgz", "mar"

frankfliu:
I don't think we should support .tgz format in model archiver. It cause confusion for customer, and MMS won't be able directly consume .tgz file.
And tar file is linux specific, windows user won't be able to use it.

model-archiver/model_archiver/arg_parser.py
ddavydenko: I really think we should go ahead with meaningful values here rather than magical ".mar" or ".tgz". These don't make sense to users on their own without some deeper knowledge of what is the difference. I suggest to go with choices of "cloud", "local" and "aws-sagemaker" with:
- 'cloud' resolving to generate ".mar" file, DEFAULT value
- 'local' resolving to generate only MANIFEST file, without archiving anything
- 'sagemaker' resolving to generate ".tar.gz" file

Also, I suggest to rename this param to "--hosting-at" or something similar. I think that would be more clear to user what this parameter is for (it is not as important to user what file is generated as a result, it is more important for user to know where it meant to be placed for inference to work).

lupesko:
@ddavydenko I read your suggestions for cloud/local/aws-sagemaker and have some concerns about them: 
(1) I find differentiating cloud from aws-sagemaker confusing, since sagemaker is a cloud-based service. 
(2) users may want to use .mar file for local as well, it is useful to have a single file to move around rather than a whole folder, so local being manifest only is restrictive in that respect.
(3) users may want to load .mar files over on-prem hosts, not just "cloud", so the differentiation feels misleading.
I tend to agree with @frankfliu comment that supporting .mar an .tgz in the export tool is confusing, and restricts cross platform support.

model-archiver/model_archiver/arg_parser.py
vdantu: I had modified the options to Amazon Sagemaker (I think thats how the service is marketed) and Native MMS. The default value is "Native MMS". I thought these options were very clear to users. Wouldn't "None" be a better option than "Local"? All these artifacts can be hosted locally,  the only difference is in the "archive-format". "None" seems to put across the message to the users users that the tools wouldn't be archiving the model artifacts to any format, more clearly.

ddavydenko: Ok, these make sense to me. I would only suggest to update "None" to "NoCompression" to be more explicit to user that there will be no archive created as a result of running the tool.

ddavydenko:
@lupesko , I agree with your comments re: naming of options. I think @vdantu has addressed them reasonably. As for supporting different options at all - we are aiming at shaping MMS to support multiple hosting solutions with pre-canned experience. More detailed: we are favoring an approach where user don't need to know which format is needed for sagemaker vs what is the native format for MMS itself. All they need to do is to specify what is the format of the output artifact for model-archiver and that is all they need to specify. Alternative would be for user to dive deep into each hosting solution (aka sagemaker) requirements and realize which archiver format it needs and only with such knowledge be able to use model-archiver correctly. This this PR we are trying to render user free of such necessity while still keeping model-archiver flexible for different hosting solutions, including native MMS format.

vdantu:
@frankfliu : Are we sure .tar.gz is not usable on windows? 

I will get a windows machine and test this link https://wiki.haskell.org/How_to_unpack_a_tar_file_in_Windows and https://sukhbinder.wordpress.com/2014/03/06/untar-a-tar-file-with-python/ 

But they seem to suggest that we can use a .tar.gz on windows as well.  With Java it seems like an additional dependency, but seems like its doable on windows as well. 
https://stackoverflow.com/questions/7128171/how-to-compress-decompress-tar-gz-files-in-java

vdantu:
@lupesko @ddavydenko @frankfliu @vrakesh : 
I made changes to the PR to add options "tgz" and "default" as per suggestions. Since this is a blocker for the release, I separated the task of consuming these artifacts and creation of "no-archive" as a backlog. I will work on getting those changes in by next release. 

It would be great to get a review on this PR.

vdantu:
@ddavydenko : Added notes in README file.

model-archiver/model_archiver/tests/unit_tests/test_model_packaging.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, **kwargs):
     export_path = '/Users/ghaipiyu/'
```
ddavydenko: I know this is not part of this PR, but this export path seems to be user specific. Can we clean it up and out into tmp folder so that it won't be in /Users folder? Or, alternatively, change it to current user's home (~/)?

model-archiver/model_archiver/tests/unit_tests/test_model_packaging.py
```diff
@@ -52,11 +53,11 @@ def test_gen_model_archive(self, patches):
...
-        patches.export_utils.check_mar_already_exists.return_value = '/Users/ghaipiyu/'
-        patches.export_utils.check_custom_model_types.return_value = '/Users/ghaipiyu', ['a.txt', 'b.txt']
+        patches.export_utils.check_mar_already_exists.return_value = '/Users/dummyUser/'
+        patches.export_utils.check_custom_model_types.return_value = '/Users/dummyUser', ['a.txt', 'b.txt']
         patches.export_utils.zip.return_value = None
 
         package_model(self.args, ModelExportUtils.generate_manifest_json(self.args))
         patches.export_utils.validate_inputs.assert_called()
-        patches.export_utils.zip.assert_called()
+        patches.export_utils.archive.assert_called()
```
piyushghai: @vdantu Can you add a test to check the code path with .tgz as the extension ? 
Right now this test checks the "default" code path (which now has branches in it due to multiple formats)

vdantu: I hadn't added the test because these tests don't seem to be testing the actual output. But you do have a valid point. I am going to add similar test for now. Will put something in the backlog to update some of the tests.

vdantu:
Will address adding integ tests for model archiver with https://github.com/awslabs/mxnet-model-server/issues/786

PR #777, opened 2019-03-19
Updates for initialize() description

ddavydenko:
Updated doc to spell out when initialize() is called and to provide explicit ideas on when it might be useful.

## Description of changes:
Updated doc to explain when initialize() is called.
## Testing done: N/A


docs/custom_service.md
```diff
@@ -106,6 +106,8 @@ def handle(data, context):
...
 ```handle()``` method itself. The ```initialize()``` method is used to initialize the model at load time, so after first time, the service need not be re-initialized in the the life cycle of the relevant worker.
  We recommend using a ```initialize()``` method, avoid initialization at prediction time.
+ 
+ This entry point is engaged in two cases: (1) when MMS gets a ```PUT /models``` request or (2) when MMS gets ```POST /predictions``` request. (1) is used to scale up workers for model that has been already registered, e.g. scale model up. (2) is used as standard way to run inference against models. (1) is also known as model load time, and that is where you would normally want to put code for model initialization. _Note_: (1) will also be engaged if during MMS startup (```mxnet-model-server --start```) you provide model(s) to load OR if when MMS receives a ```POST /models``` call with `initial-workers` option as that call is used to update model state within MMS and could result in adjustment of scale for the model, e.g. decrease or increase amount of its workers. You can find out mode about these and other MMS APIs in [MMS Management API](./management_api.md).
```
vdantu: These are not shown as points.

ddavydenko: I meant them to not be shown as points, I just wanted to enumerate them so I can refer to them by indices rather than having to re-spell in words over and over.

docs/custom_service.md
```diff
@@ -106,6 +106,8 @@ def handle(data, context):
...
 ```handle()``` method itself. The ```initialize()``` method is used to initialize the model at load time, so after first time, the service need not be re-initialized in the the life cycle of the relevant worker.
  We recommend using a ```initialize()``` method, avoid initialization at prediction time.
+ 
+ This entry point is engaged in two cases: (1) when MMS is asked to scale model up or down (it is done either via a ```PUT /models``` request or a ```POST /models``` request with `initial-workers` option or if during MMS startup (```mxnet-model-server --start```) you provide model(s) to load) or (2) when MMS gets ```POST /predictions``` request. (1) is used to scale up/down workers for model that has been already registered, e.g. scale model up or down. (2) is used as a standard way to run inference against models. (1) is also known as model load time, and that is where you would normally want to put code for model initialization. You can find out mode about these and other MMS APIs in [MMS Management API](./management_api.md).
```
vdantu:
```suggestion
 This entry point is engaged in two cases: (1) when MMS is asked to scale a model up, to increase the number of backend workers (it is done either via a ```PUT /models/{model_name}``` request or a ```POST /models``` request with `initial-workers` option or during MMS startup when you use `--models` option (```mxnet-model-server --start --models {model_name=model.mar}```), ie., you provide model(s) to load) or (2) when MMS gets a ```POST /predictions/{model_name}``` request. (1) is used to scale-up or scale-down workers for a model. (2) is used as a standard way to run inference against a model. (1) is also known as model load time, and that is where you would normally want to put code for model initialization. You can find out mode about these and other MMS APIs in [MMS Management API](./management_api.md) and [MMS Inference API](./inference_api.md)
```

PR #778, opened 2019-03-31
Batch settings in archive

### LINKED ISSUE #770: Batch Configuration In Model Archive
Follow up to https://github.com/awslabs/mxnet-model-server/issues/732#issuecomment-470214725 and https://github.com/awslabs/mxnet-model-server/issues/732#issuecomment-470224017

 #### Short Background
Currently, there are 2 ways to register models:
- POSTing the model to the management API
- Writing the model to the model repository on disk (say, `/opt/models`)

#### Problem
Some model settings, such as batching configuration, are only configurable through the register API call, but not through the disk repository.

 #### Disk Repository Use Case
Containerized applications should optimally be stateless. This makes them easier to operate and orchestrate. When we manually register models through the management API, the container is no longer stateless as it requires further configuration upon booting. If the container were to crash, it would boot into an unusable state and we'll have to reconfigure it (any model has to be POSTed again with it's configuration).

The problem becomes more severe when using an orchestrator, such as Kubernetes, which creates, relocates and evicts containers on the fly. The created containers would boot into an "unhealthy" state every time until configured.

In contrast, when a container reads the models from a predetermined location (the disk repository), it boots directly into the required state without any further, manual configuration.

#### Proposition
Having both the management API and the model repository is great. However, some parameters (most noticeably, the batching configuration) are only available through the management API. It would be great if we could bring the two loading methods to feature parity to support use cases similar to the one I described above.

Embedding the additional model settings in the model archive seems like an interesting option, as it will make the settings available to MMS when the model is loaded from the disk.

erandagan:
## Issue #, if available:
Closes #770 
### Goal
Add the ability to configure defaults for `batch_size` and `max_batch_delay` inside the model manifest. See #770 for more details.

### Overview of changes
- The CLI now exposes new optional flags, `batch-size` and `max-batch-delay`, with the default values set to 100 and 1, respectively. These defaults match the current MMS version's defaults.
- Added the new fields to the Java `Manifest.Model` class.
- Changed the `ModelManager.registerModel` method signature:
Previously, the `ModelManager` exposed a `registerModel` method which would receive a URL and nullable settings to override in the model. The manager would then setup a ModelArchive, override the settings in it, and then apply it. However, this made it difficult to override specific settings, as you'll have to chain `nulls` and zeros until you got to the value you want to change.

This PR reverses the responsibility - the `ModelManager`'s sole responsibility is to receive archives and properly register them. Configuring the archive is up to the caller. 

### Compatibility
This PR is fully backward compatible.
- The previous default batch settings (size=1, delay=100) are now applied as the default CLI arguments. 
- For compatibility with previously generated archives, the default values are also present in the Java side as GSON's defaults.

## Testing
- Added a new testing archive, `noop-batch`. It's simply `noop` with  batch settings in its manifest.
- Added new unit tests for the model server to cover this feature:
  - Test registering a model that has built in batch settings, ensure they are applied.
  - Test registering a model that has no built in batch settings, and provide the batch settings in the API call. Ensure the batch settings are applied.
  - Test registering a model that has built in batch settings, and provide the batch settings in the API call as overrides. Ensure the batch settings are overridden.

## Testing done:
- Ran local CI tests (with the newly added unit tests)


frontend/server/src/test/java/com/amazonaws/ml/mms/ModelServerTest.java
```diff
@@ -273,70 +281,112 @@ private void testLoadModelWithInitialWorkers(Channel channel) throws Interrupted
...
+                new DefaultFullHttpRequest(
+                        HttpVersion.HTTP_1_1, HttpMethod.GET, "/models/noop_v0.1");
+        channel.writeAndFlush(describeReq);
+        latch.await();
+
+        DescribeModelResponse describeResp = describeModel(channel, "noop_v0.1");
+
+        Assert.assertEquals(256, describeResp.getBatchSize());
+        Assert.assertEquals(500, describeResp.getMaxBatchDelay());
     }
```
erandagan: The diff here is sort of messed up for some reason. These methods weren't touched, except for `testUnregisterModel` and `testDescribeModel` - both of which logics has been extracted for reuse in other tests.

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
```diff
@@ -115,19 +115,19 @@ private void initModelStore() {
...
-                return;
-            }
+        String modelStore = configManager.getModelStore();
 
-            File modelStoreDir = new File(modelStore);
-            if (!modelStoreDir.exists()) {
-                logger.warn("Model store path is not found: {}", modelStore);
-                return;
-            }
+        if (modelStore == null) {
```
erandagan: Out of this PR's scope, but I wonder if we should actually throw an exception in this case as the modelStore is being relied on in several other places. It might make sense to crash MMS in case the store isn't set, rather than simply warning against it and aborting the model registration.

erandagan:
@vdantu @frankfliu @ddavydenko Hey, could one of you please take a look? Thanks!

vdantu:
Will review the code by tonight

vdantu:
There are quite a few files and changes you made. I will try to go through all the changes. Just curious, why have so many commits for a single change? Do you intend to make changes other than the "batch-configuration" coming in from MANIFEST?

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
```diff
@@ -115,19 +115,19 @@ private void initModelStore() {
 
         ModelManager modelManager = ModelManager.getInstance();
         int workers = configManager.getDefaultWorkers();
-        if ("ALL".equalsIgnoreCase(loadModels)) {
-            String modelStore = configManager.getModelStore();
-            if (modelStore == null) {
-                logger.warn("Model store is not configured.");
-                return;
-            }
+        String modelStore = configManager.getModelStore();
```
vdantu: Why should we have this change for modelStore?

frontend/server/src/main/java/com/amazonaws/ml/mms/ModelServer.java
```diff
@@ -210,7 +208,7 @@ public ChannelFuture initializeServer(
         if (connector.isSsl()) {
             sslCtx = configManager.getSslContext();
         }
-        b.childHandler(new ServerInitializer(sslCtx, connector.isManagement()));
+        b.childHandler(new ServerInitializer(sslCtx, connector.isManagement(), configManager));
```
vdantu: Why are you passing the config manager around? Wouldn't ConfigManager.getInstance() work?

frontend/server/src/main/java/com/amazonaws/ml/mms/ServerInitializer.java
```diff
@@ -38,9 +39,11 @@
      * @param sslCtx null if SSL is not enabled
      * @param managementServer true to initialize a management server instead of an API Server
      */
-    public ServerInitializer(SslContext sslCtx, boolean managementServer) {
+    public ServerInitializer(
+            SslContext sslCtx, boolean managementServer, ConfigManager configManager) {
```
vdantu: Same comment as above.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -164,16 +168,13 @@ private void handleRegisterModel(ChannelHandlerContext ctx, QueryStringDecoder d
         String modelName = NettyUtils.getParameter(decoder, "model_name", null);
         String runtime = NettyUtils.getParameter(decoder, "runtime", null);
         String handler = NettyUtils.getParameter(decoder, "handler", null);
-        int batchSize = NettyUtils.getIntParameter(decoder, "batch_size", 1);
-        int maxBatchDelay = NettyUtils.getIntParameter(decoder, "max_batch_delay", 100);
+        int batchSize = NettyUtils.getIntParameter(decoder, "batch_size", 0);
```
vdantu: Why is the default 0?

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -164,16 +168,13 @@ private void handleRegisterModel(ChannelHandlerContext ctx, QueryStringDecoder d
...
+        int maxBatchDelay = NettyUtils.getIntParameter(decoder, "max_batch_delay", 0);
         int initialWorkers = NettyUtils.getIntParameter(decoder, "initial_workers", 0);
         boolean synchronous =
                 Boolean.parseBoolean(NettyUtils.getParameter(decoder, "synchronous", null));
-        int responseTimeout =
-                Integer.parseInt(NettyUtils.getParameter(decoder, "response_timeout", "-1"));
-        if (responseTimeout == -1) {
```
vdantu: Was this removed on purpose?

frankfliu:
I don't think manifest is the right place to config batch settings.
model archive manifest is purely used for describe the property of the model itself, for example, whether the model support batch or not can be a property. But batch settings are serving related. User  might use different settings in different hosting environment.

And in future, MMS might be able to use adaptive batching.
We already can configure batch during registration. What we missing is the startup configuration. I think we should try to figure out a way to config per model setting in config.properties file.

vdantu:
@frankfliu : Configuring this from config.properties seems hard since models can come in at runtime.  I believe MANIFEST is a perfect place to put in model level configurations for customers who don't have access to management APIs. If you believe this is an option forward, we could continue discussing this in https://github.com/awslabs/mxnet-model-server/issues/770 . That issue has full context of the discussion. I agree that we are discussing implementing adaptive batching in MMS, but it seems like a model level configuration as well.

vdantu:
@erandagan : Please let us know if you have any questions about the comments. If you have any further questions about the PR, you could also ping us on Slack channel. We have a very active slack channel.

erandagan:
Hey @vdantu, I'll go over the comments, reply and make the required changes starting Sunday. Thanks, have a great weekend :)

erandagan:
Let's continue the discussion on #770 and then go back to this PR once we come to an agreement on using `config.properties` vs the manifest.

PR #797, opened 2019-05-04
Added logic to be able to configure default handler through config file

vdantu:

## Issue #, if available:
Add logic to handle default service handler if MANIFEST is not found or if no handler is configured in the MANIFEST.

## Testing done:

### DIFF
--- a/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
+++ b/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
@@ -217,6 +217,11 @@ public class ModelArchive {
                 }
             }
 
+            if (manifest.getModel().getModelName() == null
+                    || manifest.getModel().getModelName().isEmpty()) {
+                manifest.getModel().setModelName(url);
+            }
+
             failed = false;
             return new ModelArchive(manifest, url, dir, extracted);
         } finally {
--- a/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
+++ b/frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
@@ -217,11 +217,6 @@ public class ModelArchive {
                 }
             }
 
-            if (manifest.getModel().getModelName() == null
-                    || manifest.getModel().getModelName().isEmpty()) {
-                manifest.getModel().setModelName(url);
-            }
-
             failed = false;
             return new ModelArchive(manifest, url, dir, extracted);
         } finally {
@@ -328,6 +323,10 @@ public class ModelArchive {
             clean();
             throw e;
         }
+    }
+
+    public String getHandler() {
+        return manifest.getModel().getHandler();
     }
 
     public Manifest getManifest() {
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
@@ -83,7 +83,7 @@ public final class ModelManager {
 
         ModelArchive archive = ModelArchive.downloadModel(configManager.getModelStore(), url);
         if (modelName == null || modelName.isEmpty()) {
-            modelName = archive.getModelName();
+            modelName = archive.getModelName().isEmpty() ? url : archive.getModelName();
         } else {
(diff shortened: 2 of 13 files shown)


frontend/server/src/test/java/com/amazonaws/ml/mms/ModelServerTest.java
```diff
@@ -467,31 +468,35 @@ private void testPredictionsValidRequestSize(Channel channel) throws Interrupted
...
+            throws InterruptedException {
         result = null;
         latch = new CountDownLatch(1);
-        HttpRequest req =
-                new DefaultFullHttpRequest(
-                        HttpVersion.HTTP_1_1,
-                        HttpMethod.POST,
-                        "/models?url=noop-v1.0-config-tests&model_name=noop-config&initial_workers=1&synchronous=true");
+        String uri =
+                "/models?url="
```
vrakesh: Nitpick, why not StringBuilder?

vdantu: I didn't choose StringBuilder because this is a finite set of strings that need to be concatenated . I think internally the compiler would convert this to a StringBuilder.

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
```diff
@@ -217,6 +217,11 @@ private static ModelArchive load(String url, File dir, boolean extracted)
                 }
             }
 
+            if (manifest.getModel().getModelName() == null
```
frankfliu: We purposely all empty model name here, to support model archive without MANIFEST.xml. The model name checking can be default in load model management API call.
And for initial model loading, we assign model name and override in the model archive anyway

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -94,6 +94,13 @@ public ModelArchive registerModel(
             archive.getManifest().getModel().setHandler(handler);
         }
 
+        if (archive.getManifest().getModel().getHandler() == null
```
frankfliu: This should be in else condition

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -83,7 +83,7 @@ public ModelArchive registerModel(
 
         ModelArchive archive = ModelArchive.downloadModel(configManager.getModelStore(), url);
         if (modelName == null || modelName.isEmpty()) {
-            modelName = archive.getModelName();
+            modelName = archive.getModelName().isEmpty() ? url : archive.getModelName();
```
frankfliu: 1. This should be an error case, we should not use url ad model name.
2. archive.getModelName() might return null.

frontend/modelarchive/src/main/java/com/amazonaws/ml/mms/archive/ModelArchive.java
vdantu: I probably need to understand this requirement for non MANIFEST model archive that was done.  I will discuss this with you offline.

vdantu: This is required as Manifest is created empty and the model-name is not set in the Manifest. We could go in one of the two routes,
1. We could add a constructor to Manifest, which initializes the Model with a model-name, which in this case is URL.
(or)
2. We could go with this change, where we set the modelName explicitly after creating the model archive.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: I added this to "else if". The reason for this "if" is that handler is given in the POST call, then it takes the highest precedence. Else, we should rely on the handler coming through the Manifest. If the MANIFEST doesn't provide one, then we should default to the "default handler".

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
vdantu: I removed this check. I think the change above makes sure that the modelName is always set.

PR #853, opened 2019-10-23
Force synch response for unregistermodel

alexwong:
## Description of changes:
Currently we are sending a response saying that a model has been unregistered and letting the Java's Process library handle actually destroying the resources which means we can have an asynchronous response w/ respect to resource handling. This change ensures we have actually destroyed the resource before sending the response.

## Testing done:

Verified locally w/ log statements (Added log statement to right before the http response is sent back and also after each worker process is actually destroyed, verify that the last thing to happen is the response is sent). Will add a test case if feasible (Java makes it pretty hard to monitor memory usage and our test infrastructure is currently at a client req/resp level).

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -149,6 +149,7 @@ management_address=http://0.0.0.0:8081
 # number_of_netty_threads=0
 # netty_client_threads=0
 # default_response_timeout=120
+# unregister_model_timeout=120
 # default_workers_per_model=0
 # job_queue_size=100
 # async_logging=false
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -162,6 +162,7 @@ Most of those properties are designed for performance tuning. Adjusting those nu
 * job_queue_size: number inference jobs that frontend will queue before backend can serve, default 100.
 * async_logging: enable asynchronous logging for higher throughput, log output may be delayed if this is enabled, default: false.
 * default_response_timeout: Timeout, in seconds, used for model's backend workers before they are deemed unresponsive and rebooted. default: 120 seconds.
+* unregister_model_timeout: Timeout, in seconds, used when handling an unregister model request when cleaning a process before it is deemed unresponsive and an error response is sent. default: 120 seconds.
 * decode_input_request: Configuration to let backend workers to decode requests, when the content type is known. 
 If this is set to "true", backend workers do "Bytearray to JSON object" conversion when the content type is "application/json" and 
 the backend workers convert "Bytearray to utf-8 string" when the Content-Type of the request is set to "text*". default: true  
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
@@ -61,6 +61,9 @@ public class HttpRequestHandler extends SimpleChannelInboundHandler<FullHttpRequ
         } catch (ConflictStatusException e) {
             logger.trace("", e);
             NettyUtils.sendError(ctx, HttpResponseStatus.CONFLICT, e);
+        } catch (RequestTimeoutException e) {
+            logger.trace("", e);
+            NettyUtils.sendError(ctx, HttpResponseStatus.REQUEST_TIMEOUT, e);
         } catch (MethodNotAllowedException e) {
             logger.trace("", e);
             NettyUtils.sendError(ctx, HttpResponseStatus.METHOD_NOT_ALLOWED, e);
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
@@ -238,10 +238,15 @@ public class ManagementRequestHandler extends HttpRequestHandlerChain {
     }
 
     private void handleUnregisterModel(ChannelHandlerContext ctx, String modelName)
-            throws ModelNotFoundException {
+            throws ModelNotFoundException, InternalServerException, RequestTimeoutException {
         ModelManager modelManager = ModelManager.getInstance();
-        if (!modelManager.unregisterModel(modelName)) {
+        HttpResponseStatus httpResponseStatus = modelManager.unregisterModel(modelName);
+        if (httpResponseStatus == HttpResponseStatus.NOT_FOUND) {
             throw new ModelNotFoundException("Model not found: " + modelName);
+        } else if (httpResponseStatus == HttpResponseStatus.INTERNAL_SERVER_ERROR) {
+            throw new InternalServerException("Interrupted while cleaning resources: " + modelName);
+        } else if (httpResponseStatus == HttpResponseStatus.REQUEST_TIMEOUT) {
+            throw new RequestTimeoutException("Timed out while cleaning resources: " + modelName);
         }
         String msg = "Model \"" + modelName + "\" unregistered";
         NettyUtils.sendJsonResponse(ctx, new StatusResponse(msg));
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import org.slf4j.Logger;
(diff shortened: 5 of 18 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -112,7 +112,19 @@ public int getNumRunningWorkers(String modelName) {
...
+                    WorkerLifeCycle lifecycle = thread.getLifeCycle();
                     thread.shutdown();
+
+                    Process workerProcess = lifecycle.getProcess();
+
+                    if (workerProcess.isAlive()) {
+                        workerProcess.destroyForcibly();
+                        try {
+                            workerProcess.waitFor();
+                        } catch (InterruptedException e) {
```
vdantu: What could cause the process wait to be interrupted? Asking just out of curiosity. 

It seems like the future is not completed. All the other returns are returning a completed future. This might cause some issues. 

This method is called in UnregisterModel (DELETE /models) as well as scale up/down models (PUT /models). You should test for both. Could you test for both to see if anything breaks by returning just the future.

alexwong: This only happens in MMS' context if another thread explcitly calls waitingThread.interrupt before or during the waitFor which never happens. I think it's probably better to complete the future w/ an error before returning here though.

vdantu:
PR failing in python3 tests.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -112,7 +116,20 @@ public int getNumRunningWorkers(String modelName) {
...
                     WorkerThread thread = threads.remove(i);
+                    WorkerLifeCycle lifecycle = thread.getLifeCycle();
                     thread.shutdown();
+
+                    Process workerProcess = lifecycle.getProcess();
+
+                    if (workerProcess.isAlive()) {
+                        workerProcess.destroyForcibly();
+                        try {
+                            workerProcess.waitFor();
```
vdantu: Look into if we can wait for some amount of time. This otherwise seems to wait for ever (until the process finishes). [waitFor reference](https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html)

vdantu:
One thing to keep in mind with this change is, the ```modelChanged()``` method is eating the InterruptedException. We can go with this for now, but good practice would be that the internal methods don't handle exceptions and bubbles them up. The top level caller method of these internal methods handle these exceptions. Two reasons for this, one is that the internal methods might not know what to do with this exception and secondly, this could provide more flexibility for the top level callers of this internal method to implement different behaviors for these exceptions.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
```diff
@@ -61,6 +61,9 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest req) {
         } catch (ConflictStatusException e) {
             logger.trace("", e);
             NettyUtils.sendError(ctx, HttpResponseStatus.CONFLICT, e);
+        } catch (GatewayTimeoutException e) {
+            logger.trace("", e);
+            NettyUtils.sendError(ctx, HttpResponseStatus.GATEWAY_TIMEOUT, e);
```
vdantu: Probably 408 is a better. Thats a TIMEOUT_ERROR.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -239,11 +243,20 @@ private void handleRegisterModel(
 
     private void handleUnregisterModel(ChannelHandlerContext ctx, String modelName)
             throws ModelNotFoundException {
+        logger.warn("Unregister request");
         ModelManager modelManager = ModelManager.getInstance();
-        if (!modelManager.unregisterModel(modelName)) {
+        // TODO AW-10/28/19: Will need to have this return an enum so we can track multiple errors.
+        // Or just return the raw httpstatus so we can bubble up
+        HttpResponseStatus httpResponseStatus = modelManager.unregisterModel(modelName);
```
vdantu: I think unregisterModel sending a HttpResponseStatus is a good change. Yes, you could probably look into simplifying the if/else conditions by either bubbling it up or just invoking NettyUtils.sendError() here itself.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/ManagementRequestHandler.java
```diff
@@ -239,11 +243,20 @@ private void handleRegisterModel(
...
+        if (httpResponseStatus == HttpResponseStatus.NOT_FOUND) {
             throw new ModelNotFoundException("Model not found: " + modelName);
+        } else if (httpResponseStatus == HttpResponseStatus.INTERNAL_SERVER_ERROR) {
+            throw new InternalServerException("Interrupted while cleaning resources" + modelName);
+        } else if (httpResponseStatus == HttpResponseStatus.GATEWAY_TIMEOUT) {
+            throw new GatewayTimeoutException("Timed out while cleaning resources" + modelName);
         }
         String msg = "Model \"" + modelName + "\" unregistered";
+        logger.warn("Unregister response");
```
vdantu: Was this added by accident?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -126,20 +126,38 @@ public ModelArchive registerModel(
...
-    public boolean unregisterModel(String modelName) {
+    public HttpResponseStatus unregisterModel(String modelName) {
         Model model = models.remove(modelName);
         if (model == null) {
             logger.warn("Model not found: " + modelName);
-            return false;
+            return HttpResponseStatus.NOT_FOUND;
         }
-
+        logger.warn("setMinWorkers(0)");
```
vdantu: Remove all logger.warn() messages.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -126,20 +126,38 @@ public ModelArchive registerModel(
...
-        model.getModelArchive().clean();
-        startupModels.remove(modelName);
-        logger.info("Model {} unregistered.", modelName);
-        return true;
+        logger.warn("modelChanged(0)");
+        CompletableFuture<HttpResponseStatus> futureStatus = wlm.modelChanged(model);
+        HttpResponseStatus httpResponseStatus = HttpResponseStatus.OK;
+        try {
+            httpResponseStatus = futureStatus.get();
+        } catch (Exception e) {
```
vdantu: You are handling blanket exception here. Not a good idea. Handle specific exceptions.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -126,20 +126,38 @@ public ModelArchive registerModel(
...
+        logger.warn("futureStatus");
+        logger.warn(futureStatus.toString());
+        logger.warn(httpResponseStatus.toString());
+        logger.warn("cleanArchive(0)");
+
+        if (httpResponseStatus == HttpResponseStatus.OK) {
+            model.getModelArchive().clean();
+            startupModels.remove(modelName);
+            logger.info("Model {} unregistered.", modelName);
+        } else {
```
vdantu: I don't think else is required. Just log the HTTP status and the method name, every time this method is called. That could be enough.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -112,7 +117,31 @@ public int getNumRunningWorkers(String modelName) {
...
+                    WorkerLifeCycle lifecycle = thread.getLifeCycle();
                     thread.shutdown();
+
+                    Process workerProcess = lifecycle.getProcess();
+
+                    if (workerProcess.isAlive()) {
+                        boolean workerDestroyed = false;
+                        workerProcess.destroyForcibly();
+                        try {
+                            workerDestroyed = workerProcess.waitFor(2L, TimeUnit.MINUTES);
```
vdantu: I thought this would be configurable?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
vdantu: I think you added future to this, so I will resolve this comment.

frontend/server/src/main/java/com/amazonaws/ml/mms/http/HttpRequestHandler.java
alexwong: Agreed, that fits better.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
alexwong: Yep, will do once I get the final commit out.

docker/README.md
```diff
@@ -149,6 +149,7 @@ management_address=http://0.0.0.0:8081
 # number_of_netty_threads=0
 # netty_client_threads=0
 # default_response_timeout=120
+# management_response_timeout=120
```
vdantu: Naming is unclear. Could you think of a better name please?

docs/configuration.md
```diff
@@ -162,6 +162,7 @@ Most of those properties are designed for performance tuning. Adjusting those nu
 * job_queue_size: number inference jobs that frontend will queue before backend can serve, default 100.
 * async_logging: enable asynchronous logging for higher throughput, log output may be delayed if this is enabled, default: false.
 * default_response_timeout: Timeout, in seconds, used for model's backend workers before they are deemed unresponsive and rebooted. default: 120 seconds.
+* management_response_timeout: Timeout, in seconds, used for model's frontend processes before they are deemed unresponsive and an error response is sent. default: 120 seconds.
```
vdantu: Make it clear that this is for loading and unregistering models

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/ModelManager.java
```diff
@@ -126,20 +127,32 @@ public ModelArchive registerModel(
...
-        startupModels.remove(modelName);
-        logger.info("Model {} unregistered.", modelName);
-        return true;
+        CompletableFuture<HttpResponseStatus> futureStatus = wlm.modelChanged(model);
+        HttpResponseStatus httpResponseStatus = HttpResponseStatus.OK;
+
+        try {
+            httpResponseStatus = futureStatus.get();
+        } catch (InterruptedException | ExecutionException e) {
+            logger.warn("Process was interrupted while cleaning resources.");
```
vdantu: Does this go into the mms_log.log with default log.properties?

PR #874, opened 2019-11-21
More refactoring for name change

alexwong:
Update code related to docker containers after creating new repo. Also handle rest of renaming.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,4 +1,4 @@
-# Multi model server Benchmarking
+# Multi Model Server Benchmarking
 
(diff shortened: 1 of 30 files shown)


alexwong:
Wait for new S3 bucket to be created to house data for docs.

vdantu:
The changes look good. 

**NOTE: Just a side note, we should also release an mxnet-model-server pip package with the deprecation warning for the next release. Otherwise customers who are currently using mxnet-model-server might never know that they should change to multi-model-server.**

alexwong:
> The changes look good.
> 
> **NOTE: Just a side note, we should also release an mxnet-model-server pip package with the deprecation warning for the next release. Otherwise customers who are currently using mxnet-model-server might never know that they should change to multi-model-server.**

Noted.

PR #900, opened 2020-02-06
Update MacOS Java installation guide in Tutorial

StrayBird-ATSH:
## Description of changes:

The existing command in the tutorial is deprecated. Please consider updating it to the latest one for avoiding confusion or misleading newcomers.

Currently using the given command will lead to the following errors:
`Error: caskroom/versions was moved. Tap homebrew/cask-versions instead.`

Please refer to [https://github.com/Homebrew/homebrew-cask-versions](https://github.com/Homebrew/homebrew-cask-versions) or [https://stackoverflow.com/questions/44632292/how-to-install-specific-java-version-using-homebrew](https://stackoverflow.com/questions/44632292/how-to-install-specific-java-version-using-homebrew).

## Testing done:

**To run CI tests on your changes refer [README.md](https://github.com/awslabs/multi-model-server/blob/master/ci/README.md)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Before proceeding further with this document, make sure you have the following p
 
     For macOS:
     ```bash
-    brew tap caskroom/versions
+    brew tap homebrew/cask-versions
     brew update
     brew cask install java8
     ```
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Before proceeding further with this document, make sure you have the following p
     ```bash
     brew tap homebrew/cask-versions
     brew update
-    brew cask install java8
+    brew cask install adoptopenjdk8
     ```
 
 ### Installing Multi Model Server with pip


aaronmarkham:
I think there's a new requirement from Oracle that you have to sign a license agreement, and download the software from them. 
This installation path is a bit... convoluted, so it might be better to just say: You need x version of Java (JRE/JDK?) and put that in the prerequisites. Then delete these lines about installing Java via brew.

README.md
```diff
@@ -47,7 +47,7 @@ Before proceeding further with this document, make sure you have the following p
 
     For macOS:
     ```bash
-    brew tap caskroom/versions
+    brew tap homebrew/cask-versions
     brew update
     brew cask install java8
```
frankfliu: brew cask install adoptopenjdk8

frankfliu:
> I think there's a new requirement from Oracle that you have to sign a license agreement, and download the software from them.
> This installation path is a bit... convoluted, so it might be better to just say: You need x version of Java (JRE/JDK?) and put that in the prerequisites. Then delete these lines about installing Java via brew.

We should install openjdk instead.

PR #911, opened 2020-04-26
Update Docker file to MXNet 1.6

hassonofer:
## Issue #, if available:
No issue is open...

## Description of changes:
Update MXNet version to 1.6 on 2 Dockerfiles CPU and GPU.
The PIP package already suggest the latest version of MXNet.
This only to keep the containers and non Docker installations the same.


hassonofer:
As this is my first contribution here, I didn't want to make a mess :)

But I think that the Dockerfile would be simpler if it pulled from Python instead of Ubuntu.

PR #915, opened 2020-05-13
Clean up resources after worker thread is terminated

maaquib:
## Description of changes:
* Terminate err and out ReaderThreads after its corresponding WorkerThread has been terminated
* Clean up FDs associated with a terminated WorkerThread
* Pins [pytest](https://github.com/pytest-dev/pytest/issues/3518) version in CI test to avoid error `fixture is being applied more than once to the same function`

## Testing done (Ubuntu18.04):
* Stress tested for 450000 iterations of model load and unload
* CI [tests](https://github.com/awslabs/multi-model-server/blob/master/ci/README.md) successful

### DIFF
--- a/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
+++ b/frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
@@ -113,6 +113,9 @@ public class WorkLoadManager {
             if (minWorker == 0) {
                 threads = workers.remove(model.getModelName());
                 if (threads == null) {
+                    if (maxWorker == 0) {
+                        return shutdownServerThread(model, future);
+                    }
                     future.complete(HttpResponseStatus.OK);
                     return future;
                 }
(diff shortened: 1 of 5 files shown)


frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -129,37 +132,40 @@ public int getNumRunningWorkers(String modelName) {
...
+                logger.warn(
+                        "WorkerThread interrupted during waitFor, possible asynch resource cleanup.");
+                future.complete(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+            }
+            if (!workerDestroyed) {
+                logger.warn("WorkerThread timed out while cleaning, please resend request.");
+                future.complete(HttpResponseStatus.REQUEST_TIMEOUT);
+            }
+        }
+        future.complete(HttpResponseStatus.OK);
```
vdantu, line 165: Seems like `future` is always completed with `HttpResponseStatus.OK`.. That doesn't look correct..  Shouldn't we return the `future` whenever its completed?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerLifeCycle.java
```diff
@@ -210,10 +228,14 @@ public ReaderThread(String name, InputStream is, boolean error, WorkerLifeCycle
...
 
+        public void terminate() {
+            isRunning = false;
+        }
+
         @Override
         public void run() {
             try (Scanner scanner = new Scanner(is, StandardCharsets.UTF_8.name())) {
-                while (scanner.hasNext()) {
+                while (isRunning && scanner.hasNext()) {
```
vdantu, line 238: Adding flag looks a little hacky.

frontend/server/src/test/java/com/amazonaws/ml/mms/ModelServerTest.java
```diff
@@ -1312,7 +1312,7 @@ private void testLoggingUnload(Channel inferChannel, Channel mgmtChannel)
         Scanner logscanner = new Scanner(logfile, "UTF-8");
         while (logscanner.hasNextLine()) {
             String line = logscanner.nextLine();
-            if (line.contains("LoggingService exit")) {
+            if (line.contains("Model logging unregistered")) {
```
vdantu, line 1315: Whats this change? I don't see corresponding change in the source file. Curious to know how this log line gets into the logfile.

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -381,6 +384,22 @@ public void shutdown() {
...
             backendChannel.close();
         }
+        if (this.serverThread && this.connector != null) {
+            logger.debug("Cleaning connector socket");
+            this.connector.clean();
+        }
+        logger.debug("Terminating IOStreams for worker thread shutdown");
+        lifeCycle.terminateIOStreams();
+        try {
+            if (out != null) {
```
vdantu, line 394: When would we have out or error equal to null? Shoudln't they always be created?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkerThread.java
```diff
@@ -381,6 +384,22 @@ public void shutdown() {
...
+                out.close();
+            }
+            if (err != null) {
+                err.close();
+            }
+        } catch (IOException e) {
+            logger.error("Failed to close IO file handles", e);
```
vdantu, line 401: Whats the cleanup process if we fail to close the IOs?

frontend/server/src/main/java/com/amazonaws/ml/mms/wlm/WorkLoadManager.java
```diff
@@ -113,6 +113,9 @@ public int getNumRunningWorkers(String modelName) {
             if (minWorker == 0) {
                 threads = workers.remove(model.getModelName());
                 if (threads == null) {
+                    if (maxWorker == 0) {
```
vdantu, line 116: I am not sure if `maxWorker==0` check is correct. Could you test the following sequence:1
1. Register a model (no initial workers)
2. Scale up the workers (maybe to 2 workers)
3. Scale down the workers to 0.
4. Scale up the workers again (maybe to 2 workers)..
Minworkers and Maxworkers are always the same number. So, when we scale down to 0, according to this change we would be remvoing the server thread. When will the server thread again be created? Its initially created during registration of the model.

maaquib: This was failing even before the fix. If you initialiaze a model with workers > 0, scale down to 0 and scale up again, we get an exception with serverthread being null. So essentially, there is no change in behaviour.
I can create a separate issue to track this and will send out another PR.

maaquib: Created a separate issue to fix this bug https://github.com/awslabs/multi-model-server/issues/916

vdantu: Thanks for raising the issue. Its definitely a bug that needs fixing. So currently scaledown to 0 and scaleup is broken. Only work around would be `DELETE /models` and `POST /models`.

